PR #7, opened 2015-02-12
OS X Update

ahwagner:
A couple of updates to the dgi-db code for native operability with OS X 10.10.

therubyracer and libv8 needed to be updated for compatibility with gcc 4.8 (see https://github.com/cowboyd/libv8/pull/95)--these updates are implemented in this pull request. These updates do not impact the function of the site or cause any tests to fail in a linux test environment.

In addition, the dgidb:load_local troubleshooting step in the OS X install instructions refers to the incorrect location for data.sql. This is addressed below.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@
 # Ignore bundler config
 /.bundle
 
+# Ignore generated data directory
+/data
+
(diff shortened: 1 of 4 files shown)


acoffman:
Thanks for these!

PR #9, opened 2015-02-16
OS X update

ahwagner:
Update OS X install directions and db_bootstrapping.rake to overcome some difficulties observed in setting up a new dev environment on OS X 10.10


acoffman:
Awesome, looks good!

PR #11, opened 2015-04-16
include header flag in importer class

ahwagner:
Headers are expected in .tsv importers, though this is not explicitly stated in the documentation. Adding code to allow import files without headers, in preparation for the new importer documentation.


malachig:
+1

acoffman:
+1 - looks good to me!

ahwagner:
Thanks guys! Merging #11

PR #29, opened 2015-08-17
Update GuideToPharmacologyInteractions importer. Update drug grouper.

ahwagner:
Add in nomenclature-sensitive drug grouping for numeric names. Update GTP Interactions to include a "sanitized", html-free version of `primary_name` as a `drug_alias`.


ahwagner:
Good call, thanks @acoffman. I've switched over to the strip_tags method from `Sanitize_Helper`.

PR #30, opened 2015-08-20
DrugBank Updates

ahwagner:
This branch off of PR #29 provides an update to the DrugBank interactions resource. 

This update includes:
* An iPython notebook to illustrate scripting for automatic updates
* Details for the download, parsing, and writing to tsv for the DrugBank resource
* An updated tsv_importer module for DrugBank records


ahwagner:
There are more updates coming in the next few days, that I'd like to build off of this. @acoffman @malachig @nspies13 could I get a couple of you to look this over?

nspies13:
You got it! What do you need me to do?

Nick

ahwagner:
Run the [python3 update script](https://github.com/genome/dgi-db/pull/30/files#diff-b500db89f58a14af4215da34db3702a7) to generate the tsv (it requires the BeautifulSoup and wget packages), and then load the tsv in your local instance with the (DrugBank TSV importer)[https://github.com/genome/dgi-db/pull/30/files#diff-bd088bdc072462e53c5edb9b8f9cd381] using the (DrugGrouper)[https://github.com/genome/dgi-db/pull/30/files#diff-3e21362cf36a4ed89fc87586d78b5094]:

`rake dgidb:import:drug_bank['lib/genome/updaters/data/DrugBank.tsv',true]`

Then launch a local instance and make sure everything checks out. +1 if it works as expected and there aren't any issues with the code.

obigriffith:
Some comments and an error: 

Comments:
1) Check drugbank version before downloading large files?
2) Add some error checking to deal with possibility of incomplete downloads?
3) I don't recall the details but I thought that we had some logic to identify anti-neoplastic drugs from drugbank?

I got a current local version of dgidb up and running, installed latest python3 (and wget and bs4) and ran the updater. It worked until the following. Where is "data/old.drugbank.xml" meant to come from?  I don't think the data folder where it is looking even existed until I ran the script. Is it expected that the folder already exists and an existing copy of a previous drugbank xml be copied there?

Obis-MacBook-Air:updaters ogriffit$ python3 updateDrugBank.py 
Downloading DrugBank XML...
100% [........................................................................] 47361250 / 47361250
Extracting DrugBank XML...
Downloading Entrez Accessions...
File 1:
100% [......................................................................] 721371260 / 721371260
File 2:
100% [......................................................................] 217882432 / 217882432
Extracting Entrez Accessions...
File 1...
File 2...
Checking DrugBank Version...
Parsing Entrez...
Parsing DrugBank XML...
Traceback (most recent call last):
  File "updateDrugBank.py", line 116, in <module>
    tree = ET.parse('data/old.drugbank.xml')
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/xml/etree/ElementTree.py", line 1187, in parse
    tree.parse(source, parser)
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/xml/etree/ElementTree.py", line 587, in parse
    source = open(source, "rb")
FileNotFoundError: [Errno 2] No such file or directory: 'data/old.drugbank.xml'

obigriffith:
I grabbed an old version of drugbank.xml and put in place to avoid the error above. But, I keep getting a tsv file with header only. Not sure what is going wrong...

ahwagner:
@obigriffith thanks for reviewing this. I just uploaded a big commit that cleans up the `get_drugbank.py` (previously 1updateDrugBank.py`) file.

You should remove the data directory completely (as a new user would not have one) and test again.

I've also added code to address version checking, so re-downloading now won't occur if the version hasn't changed.

I'm looking into the incomplete downloads comment, as well as the anti-neoplastic logic.

obigriffith:
OK. Trying the new version. Does this mean you are pulling in an old.drugbank.xml from somewhere automatically now?  Also, thinking back it might be that the antineoplastic stuff was handled in the importer for drugbank or somewhere else on the back end and simply used fields (drug type, etc) that were pulled in at import?

ahwagner:
No, old.drugbank.xml was an artifact of my testing—one that didn’t get removed. The new version just compares version numbers scraped off the web—negating the need to download a new version to compare against.

As for the antineoplastic stuff, it is the drug type field, I’m just trying to figure out where the code is that makes these entries. There are 492 entries from DrugBankInteractions that contain the word ‘antineoplastic’ in the drug type field. I’ll keep digging.

> On Aug 23, 2015, at 10:57 PM, Obi Griffith <notifications@github.com> wrote:
> 
> OK. Trying the new version. Does this mean you are pulling in an old.drugbank.xml from somewhere automatically now? Also, thinking back it might be that the antineoplastic stuff was handled in the importer for drugbank or somewhere else on the back end and simply used fields (drug type, etc) that were pulled in at import?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/genome/dgi-db/pull/30#issuecomment-134016504>.
> 



____

ahwagner:
Found it. We identify antineoplastic drugs in the [Drug Type Normalizer](https://github.com/genome/dgi-db/blob/master/lib/genome/normalizers/drug_claim_type.rb#L5). All that is required is a name match to an attribute. To this end, I have a very small commit (868833d2d81d77bd95d9d04730a536051406faf8) that applies a lower-case operation on the `drug_categories` field.

obigriffith:
Cool. My import attempt completely successfully and at quick glance things look as they should. Only problem I identified is that external links to drugbank (Drugs and molecules) are broken. Here are two examples:
http://www.drugbank.ca/drugs//drugs/DB01268
http://www.drugbank.ca/drugs//molecules/BE0000029/?as=target

I think the new linking style should be:
http://www.drugbank.ca/drugs/DB01268
http://www.drugbank.ca/biodb/bio_entities/BE0000029

acoffman:
This looks good to me! I also ran it to completion successfully. I will dig around a little more this afternoon, but it looks good to me so far!

ahwagner:
@obigriffith, the linkouts should work as expected now. Thanks @acoffman for the review.

obigriffith:
The drugbank import looks as it should now as far as I can tell.

Issue #23, opened 2015-06-16
Automatic Updates


ahwagner:
Add functionality for automatic updating of some resources:
- GO (**DONE** -- Added in #36)
- Guide To Pharmacology 
- Therapeutics Target Database
- My Cancer Genome (maybe? see [licensing page](http://www.mycancergenome.org/about/api-licensing/))
- PharmGKB
- DrugBank (**DONE** -- Added in #30)
- Clearity Foundation (Need to explore if this is allowable)
- Entrez (**DONE** -- Added in #33)
- Ensembl
- PubChem (**DONE** -- Added in #41)
- DoCM (**DONE** -- Added in #38)
- CIViC (**DONE** -- Added in #32)

ahwagner:
This is covered by #148

PR #33, opened 2015-08-25
Entrez updates

ahwagner:
Entrez auto-updates for #23 .

### DIFF
--- a/lib/genome/importers/civic/civic_tsv_importer.rb
+++ b/lib/genome/importers/civic/civic_tsv_importer.rb
@@ -20,6 +20,7 @@ module Genome
           source_db_version: get_version,
           source_type_id: DataModel::SourceType.INTERACTION,
           source_db_name: 'CIViC',
+          source_trust_level_id: DataModel::SourceTrustLevel.EXPERT_CURATED,
           full_name: 'CIViC: Clinical Interpretations of Variants in Cancer'
         }
       end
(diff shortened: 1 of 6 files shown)


ahwagner:
For anyone reviewing this:

```shell
cd $PROJECT_ROOT/lib/genome/updaters/
python3 get_entrez.py

cd $PROJECT_ROOT
rake dgidb:import:entrez[lib/genome/updaters/data/entrez_genes.tsv,true]
```

obigriffith:
So far everything looks as it should except the "inter-gene interactions" aka related genes are not being imported from Entrez. Compare the source summary for Entrez in this version vs public site. I don't recall if this was handled by a separate import or not but in any case it should be done at the same time. 

screenshot 2015-08-25 14 04 23
screenshot 2015-08-25 14 05 34

acoffman:
I believe there is a separate importer for Gene-Gene interactions under `lib/genome/importers/entrez/entrez_gene_pathway_importer.rb`

obigriffith:
What do you guys think?  Maybe this should just be incorporated into a single automated import/update for Entrez?

ahwagner:
@obigriffith I think it should be. The short-term solution is to just hard-code the update cron to run the `EntrezGenePathwayImporter` immediately after the `EntrezGenesImporter` when Entrez is updated.

For now, Entrez inter-gene interactions can be updated as follows:

```shell
cd $PROJECT_ROOT/lib/genome/updaters/
rm data/version
python3 get_entrez.py

cd $PROJECT_ROOT
rake dgidb:import:entrez_pathway[lib/genome/updaters/data/interactions.human]
rake dgidb:normalize:populate_source_counters
```

image

PR #36, opened 2015-08-26
Add GO updater.

ahwagner:
To test:

```shell
cd $PROJECT_ROOT/lib/genome/updaters/
python3 get_go.py

cd $PROJECT_ROOT
rake dgidb:import:go[lib/genome/updaters/data/go.human.tsv,true]
```

### DIFF
--- a/lib/genome/importers/go/go_row.rb
+++ b/lib/genome/importers/go/go_row.rb
@@ -3,8 +3,8 @@ module Genome
   module Importers
     module Go
       class GoRow < Genome::Importers::DelimitedRow
-        attribute :uniprotkb_id
         attribute :gene_name
+        attribute :uniprot_ids, Array, delimiter: '|'
         attribute :gene_category
 
         def valid?(opts = {})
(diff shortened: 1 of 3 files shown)


ahwagner:
I forgot to mention that `get_go.py` will require the `selenium` python package (`pip3 install selenium`) and the [PhantomJS executable](https://github.com/eugene1g/phantomjs/releases) somewhere in your PATH.

ahwagner:
Hold off on reviewing this one. We're currently investigating why this new process returns fewer results than the original one through the (now defunct) GO AMIGO API. The [current state](http://wiki.geneontology.org/index.php/AmiGO_2_Manual:_Services) of the AMIGO 2.0 API is 'under proposal'. I'm investigating if the [EBI QuickGO API](https://www.ebi.ac.uk/QuickGO/WebServices.html) can be used in lieu of this.

ahwagner:
The issue has to do with parent-child relationships. My approach had me searching for matches based on GO term, but didn't consider GO terms of descendants. Fortunately, the EBI API returns both, greatly easing the burden. Splitting the query by parent GO ID should make it go relatively quick. Unless I have time left at the end of the week, I will hold off on that update for after the paper submission, however.

ahwagner:
This is finished.

```shell
cd $PROJECT_ROOT/lib/genome/updaters/
python3 get_go.py

cd $PROJECT_ROOT
rake dgidb:import:go[lib/genome/updaters/data/go.human.tsv,true]
```

image

PR #41, opened 2015-08-27
Initial pubchem updater

acoffman:
@ahwagner  had asked me to take a stab at bringing the old PubChem importer over from the Genome repo. Here's a first draft - I'm chatting with him now about it.

### DIFF
--- a/lib/genome/updaters/get_pubchem.rb
+++ b/lib/genome/updaters/get_pubchem.rb
@@ -15,7 +15,8 @@ module Genome
         puts 'Downloading PubChem file...'
         getter.download_file(tmpfile.path)
         puts 'Processing PubChem file...'
-        getter.process_pubchem_file(tmpfile.open) ensure
+        getter.process_pubchem_file(tmpfile.open)
+      ensure
         puts 'Cleaning up tmpfile...'
         tmpfile.close
         tmpfile.unlink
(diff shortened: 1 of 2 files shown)


acoffman:
This works via the same general algorithm as the [perl version](https://github.com/genome/genome/blob/master/lib/perl/Genome/DruggableGene/Command/Import/Pubchem.pm):

* Delete existing PubChem data in the DGIdb database
* Download the gzipped PubChem synonyms file
* Iterate the PubChem file, looking for drugs for which we have interactions that match something in PubChem
* For matches, create a new DrugClaim with the primary PubChem compound name and import all the remaining PubChem names for the same compound as synonyms
* Import the PubChem CID as an alias for the DrugClaim

It can be invoked with `rake dgidb:import:pubchem` and takes several minutes to run to completion (including the download) as the file is > 100,000,000 lines.

ahwagner:
The problem may be [here](https://github.com/genome/dgi-db/blob/master/lib/genome/groupers/drug_grouper.rb#L23-L27). If your new alias is mixed case, it wouldn't match the regex, and the drug grouper won't know what to match to.

ahwagner:
Hey @acoffman, is 3eed6c3 the most recent commit to this branch?

acoffman:
@ahwagner Yup! That is a version of the importer that resulted in the numbers I sent you on Friday.

ahwagner:
Great. Just checking. +1

> On Aug 31, 2015, at 10:15 AM, Adam Coffman <notifications@github.com> wrote:
> 
> @ahwagner <https://github.com/ahwagner> Yup! That is a version of the importer that resulted in the numbers I sent you on Friday.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/genome/dgi-db/pull/41#issuecomment-136402813>.
> 



____

Issue #21, opened 2015-06-16
Search by Drug


ahwagner:
Currently we support searching by a list of genes. We should improve this by also allowing search by a list of drugs.

ahwagner:
Added in 2.0 release.

PR #43, opened 2015-08-30
Add drug search

ahwagner:
See issue #21.


ahwagner:
@obigriffith @malachig This is a big update, please give it a test when you have a chance--I'll go ahead and merge with a +1.

obigriffith:
+1

Overall it seems to be working well enough to go live for reviewers. 

A few minor issues:
* I notice a small graphical glitch when flipping between 'gene' and 'drug' search where it briefly shows a loading icon of some kind. Its a minor annoyance. 
* The "Search terms with no matches" box is not as wide as the "primary results" box. 
* The subtitle for the drug-driven interaction search results is "drug interactions for your genes". That kind of makes it sound like you searched for genes. Maybe, now that there are both kinds of searches we should say: 1. "interactions for your genes" and 2. "interactions for your drugs". Or, more verbosely "drug-gene interactions for your genes [drugs]"

ahwagner:
1) I never liked that loading bar, it also bothered me tremendously. It's out.
2) This is something that I'll look into. It's been this way so that it fits nicely with ambiguous results (I added new demo terms to illustrate this)--but we should investigate making it a dynamic width.
3) Good catch on the interaction result subtitle. Easy fix.

obigriffith:
Ah. Yeah. I see that now. I love how convincing sounding NOTREAL is as a drug. Now that we have the ambiguous examples in the demo list this is a very minor problem. Another solutions would be to always show all the boxes (primary, ambiguous, no matches) but just say "None" or have a blank result if no search terms fall into that category.

ahwagner:
Closes #43

PR #63, opened 2015-10-06
Closes #15. Closes #55. Add summary metric.

### LINKED ISSUE #15: By source table no longer displays correctly
We now have enough sources that the 'By Source' tab does not render nicely.  This can be seen with the demo list of gene in the interaction search page.

http://dgidb.genome.wustl.edu/interaction_search_results

There a various ways we could fix this.  Quick and dirty could be to come up with shorter names for each source that show the complete name when we mouse over them.  Or the name could wrap instead of being only on a single line.  Or we could rework this table in some other way that still achieves the same effect of having only one row per unique drug-gene interaction with some concept of the independent support for that interaction being summarized.

### LINKED ISSUE #55: Add PMIDs from sources
If a source contains publication information about an interaction, grab that information and display this on the "by source" page.

ahwagner:


ahwagner:
db issue is causing check failure. Addressing issue in #64.

PR #82, opened 2015-10-09
Fixes the `can't convert nil into String` bug.

ahwagner:
This bothered me, so I took another look at it.

It appears that any new genes we add don't have the "long name" field filled out, which was causing the error.

I'll have to do some digging to find out how this was loaded before we started adding genes for 2.0. My guess is that it was some backend magic.

@acoffman, do you know?

### DIFF
--- a/app/models/gene_category_search_result.rb
+++ b/app/models/gene_category_search_result.rb
@@ -21,7 +21,7 @@ class GeneCategorySearchResult
   end
 
   def gene_display_name
-    genes.first.long_name
+    genes.first.long_name || genes.first.name
   end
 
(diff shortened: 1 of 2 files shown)


acoffman:
I'm digging through as well and its not immediately apparent either. If I had to guess, I would have thought we were getting it from Entrez, but I sure don't see where that's happening in the code. I'll keep digging!

malachig:
I suppose this could have been something that was just done ad hoc with a file from Entrez that had a simple mapping from Entrez ID to description.  May have been something done at the last minute prior to the last submission...

PR #97, opened 2016-01-15
escape drug names with conflicting characters

ahwagner:
@acoffman if you could review this PR, I'd appreciate it. It's not very DRY, but I wasn't sure how to achieve what I wanted without adding `CGI::escape` at every call to `drug_path`. Also, I couldn't find a defined method for `drug_path`... is this one of those nifty Rails conventions I simply don't know about?

### DIFF
--- a/app/controllers/drugs_controller.rb
+++ b/app/controllers/drugs_controller.rb
@@ -1,7 +1,7 @@
 class DrugsController < ApplicationController
 
   def show
-    drug = DataModel::Drug.for_show.where('lower(drugs.name) = ?', params[:name].downcase)
+    drug = DataModel::Drug.for_show.where('lower(drugs.name) = ?', CGI::unescape(params[:name].downcase))
       .first || not_found("No drug matches name #{params[:name]}")
 
     @drug = DrugPresenter.new(drug)
--- a/app/models/interaction_search_result.rb
+++ b/app/models/interaction_search_result.rb
@@ -7,26 +7,22 @@ class InteractionSearchResult
     @search_terms = search_terms
     @type = type
     @identifiers = identifiers.uniq
-    exclude_chars = %w{/ , [ (}
     if type == 'genes'
       results = identifiers.flat_map { |g| g.gene_claims }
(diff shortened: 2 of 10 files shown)


acoffman:
Hrm, so it seems like the origin of this issue is using the drug name in the URL path. Is that correct? If so we could use a pre-generated url slug instead of the name - the I guess that would break backwards compatibility though. There's probably something we can do about that though. Perhaps support either url structure?

Yeah, the `drug_path` method comes "for free" if you name a route in `config/routes.rb`. So if there's a drugs route, you get the `drug_path` method as part of that!

ahwagner:
Yeah, that’s the problem. Example: the `CDK/CRK INHIBITOR` drug. The way I’ve coded it now still wouldn’t address somebody manually entering the URL (e.g. `dgidb.genome.wustl.edu/drugs/CDK/CRK INHIBITOR`), but rather allows for the auto-generated links to these records to go to the correct location. If we could parse incoming requests similarly, it may negate the need for a pre-generated slug. Thoughts?

> On Jan 15, 2016, at 3:43 PM, Adam Coffman <notifications@github.com> wrote:
> 
> Hrm, so it seems like the origin of this issue is using the drug name in the URL path. Is that correct? If so we could use a pre-generated url slug instead of the name - the I guess that would break backwards compatibility though. There's probably something we can do about that though. Perhaps support either url structure?
> 
> Yeah, the drug_path method comes "for free" if you name a route in config/routes.rb. So if there's a drugs route, you get the drug_path method as part of that!
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/genome/dgi-db/pull/97#issuecomment-172102716>.
> 



____

acoffman:
So, you can actually "force" that route to work correctly if that would be a better outcome. You can provide a regular expression in the route definition that will match the drug name. This will allow things like forward slashes to be valid in the URL. We're doing something similar in DoCM to allow for people to use HGVSs in the URL, even though it wouldn't be valid otherwise. You can see an example of that here: https://github.com/genome/docm/blob/master/config/routes.rb#L13

PR #114, opened 2016-06-29
Add sources to grouper

gspies:
I was unable to fix the problem with mismatched versions of rails/postgres, so I am unable to test the grouper.


ahwagner:
Rejecting PR, more recent version of this on [add-test-to-grouper](https://github.com/genome/dgi-db/compare/add-test-to-grouper) branch.

PR #116, opened 2016-09-27
Fix unmatched count

gspies:


gspies:
You can delete the non_match_val variable, I forgot to remove it.

PR #128, opened 2017-03-03
New table drug_attributes

susannasiebert:
This PR also removes the `description` column on `drug_claim_attributes`.


susannasiebert:
During the meeting we decided to add a bridge table between drug_attributes and sources. I will close this PR and work on a new one that implements that change.

PR #153, opened 2017-04-21
update contact page to 3.0 team

### LINKED ISSUE #151: Update contacts page
It really is quite out of date.  1/2 people there are no longer working at MGI.  Others would deserve credit are probably missing.  Maybe move people that are gone to the "additional conceptual contributors" section and have a separate list for the current team.

Also we currently have almost no MGI or WASHU branding anywhere on the site.  This is an unfortunate oversight.  Would be good to add that to the footer as we have on CIViC and other projects.

ahwagner:
closes #151. Since this has everyone's contact info on it, I'd like everyone to review and approve before merge. Feel free to commit changes.


malachig:
+1 looks reasonable to me.  We should change Genome Institute to McDonnell Genome Institute, throughout the entire site.

ahwagner:
Thanks @malachig. I think I got them all site-wide, but if you (or anyone else) spots a place where I missed it, please add it in.

PR #157, opened 2017-04-26
closes #69

### LINKED ISSUE #69: GO linkouts to AMIGO are broken
Fix this on the gene group page.

ahwagner:


ahwagner:
Hey @yanyangfeng, have you done a GitHub review before? This one is both straightforward and UI related, and I think it would be good for you to get some practice with it here. Let me know if you need help doing a review.

ahwagner:
Thanks for the reviews, guys. @yanyangfeng, I agree that we should be able to find claims we grabbed from that source--would you make an issue with some specific examples of where there was no match?

PR #179, opened 2017-06-05
Update to the gene import/grouping process

susannasiebert:
This PR introduces new modules for importing entries from Entrez and creating gene groups. It also updates the gene grouper to only add gene claims to the already existing groups.

As part of this PR we also did some data cleanup:
- A fresh import from Entrez was conducted, using the new importer.
- All gene claims from Entrez have been deleted. Entrez is now the canonical source for gene groups.
- After grouping the remaining claims, all gene claims from Ensembl that were not grouped and that don't have any interaction claims were deleted.

There are still a few hundred gene claims that are ungrouped and should be manually checked, updated, or deleted.
```
select gene_claims.source_id,count(*) from gene_claims left outer join interaction_claims on interaction_claims.gene_claim_id = gene_claims.id where gene_id is null and interaction_claims.gene_claim_id is null group by gene_claims.source_id;
              source_id               | count 
--------------------------------------+-------
 0b65e387-0eed-4033-b2a2-2bc4fd47911f |     5
 107f043f-5f09-4e95-8e1b-ae6a9d4375d2 |   278
 68b2e42e-be25-4ccd-b641-2df7dfdb00b9 |     2
 806c8e2b-58e0-4b72-bfc7-a70aa3eaaff4 |     4
 ff69c920-951a-448e-a376-cb7cf461692c |   440
```


susannasiebert:
@ahwagner Do you mean that the data dump needs to be updated as part of this PR to reflect the updated numbers or do you mean that the call to Genome::Normalizers::PopulateCounters needs to be part of one of the modules in this PR?

ahwagner:
@susannasiebert I was thinking the latter, although in retrospect other groupers will likely be run at the same time, and the normalizer run after. So, you can ignore that request and I'll just put it in an issue 😄.  However, I do have an ongoing review of this PR open in another tab, and there will be more requested changes (in the committed code) to review.

lib/genome/importers/entrez/new_entrez.rb
```diff
@@ -0,0 +1,115 @@
...
+          if synonym.strip == '-'
+            nil
+          else
+            create_alias(gene, synonym)
+          end
+        end
+      end
+    end
+
+    def process_ensembl(gene, xrefs)
```
ahwagner: We should consider adding all xrefs here, instead of just Ensembl.

lib/genome/importers/entrez/full_gene_import.rb
```diff
@@ -0,0 +1,59 @@
...
+            existing_alias.first
+          else
+            DataModel::GeneAlias.where('alias' => synonym, 'gene_id' => gene.id).first_or_create
+          end
+        end
+        if (existing_alias = DataModel::GeneAlias.where('lower(alias) = ? and gene_id = ?', line['Symbol'].downcase, gene.id)).any?
+          existing_alias.first
+        else
+          DataModel::GeneAlias.where('alias' => line['Symbol'], 'gene_id' => gene.id).first_or_create
+        end
```
susannasiebert: I just noticed that we don't add the resulting gene aliases to a source (i.e., Entrez). That seems like an oversight. Although we don't create a source entry for Entrez in this module at all. In fact, I wonder if this module can even just go away in favor of new_entrez. @acoffman thoughts?

acoffman: It does seem like maybe we only need the one module? I'm trying to remember why we had split it into two but can't seem to recall a good reason.... As long as `new_entrez` is capable of doing the full import, I don't see a need for a separate class.

susannasiebert: Maybe one of us should run new_entrez on an empty database to ensure that it would work on a from-scratch import. Once we check that I'm in favor of deleting the full_gene_import.

ahwagner: I'm doing this now.

ahwagner: The correct number of genes are populated, as well as gene aliases and gene alias source relationships. I'd say we're good to drop this module.

lib/genome/groupers/gene_grouper.rb
```diff
@@ -1,160 +1,111 @@
...
-
-          direct_groups << gene_claim.name if @gene_names_to_genes[gene_claim.name]
-          gene_claim.gene_claim_aliases.map(&:alias).each do |gene_claim_alias|
-            direct_groups << gene_claim_alias if @gene_names_to_genes[gene_claim_alias]
-            alt_genes = @alt_to_other[gene_claim_alias].map(&:gene_claim)
-            alt_genes.each do |alt_gene|
-              indirect_gene = alt_gene.gene
-              indirect_groups << indirect_gene.name if indirect_gene
-            end
+      def create_gene_alias(gene_id, name, source)
```
ahwagner: This is a departure from our previous direct/indirect grouping strategy.

One of the key problems with a "first match" approach like this is that it can lead to clusters of unrelated gene claims getting grouped together. Imagine if a gene claim has a mix of good and bad aliases (e.g. good: _ENSG\d+_ bad: _"N/A"_ or _"HTTP"_). It groups properly because of the good alias, but future claims group against it matching on _bad_ aliases. This could cause incorrect _good_ (but previously ungrouped) aliases to group with the wrong gene.

Previously, we handled this by requiring that a claim had only a single direct match (match to a `Gene`), and if it had no direct matches, group if it had only a single indirect match (match to a `GeneAlias`). Thus, if a bad alias got thrown in the mix, the claim would match ONLY if no other gene aliases matched it and the bad alias existed in only one other alias.

What do you think? Is that being handled here, and I'm missing it?

susannasiebert: Given this premise, the previous strategy makes sense (and can probably be equally implemented here). However, I would argue that the examples for bad aliases should not get imported in the first place since they don't seem meaningful.

susannasiebert: We could change line 42 to first check how many matched gene aliases there were and only proceed if there was one match. For the other cases, I don't think we need additional checks.

susannasiebert: I made a new commit to make the changes discussed off-ticket.

lib/genome/groupers/gene_grouper.rb
```diff
@@ -20,32 +18,39 @@ def run
       end
 
       def add_members(claims)
-        grouped_claims = Set.new()
+        grouped_claims = []
         claims.each do |gene_claim|
+          grouped_claims << group_gene_claim(gene_claim)
+        end
+        return grouped_claims.compact
```
ahwagner: slick!

Issue #173, opened 2017-05-30
Fix the source drop sql query to accomodate new db structure


ahwagner:
see https://github.com/griffithlab/dgi-db/blob/master/lib/utils/database.rb#L7

PR #180, opened 2017-06-06
Civic updater

acoffman:
This is the first of the online updaters that @susannasiebert and I worked on. It does direct database updates instead of going through the intermediate step of generating a TSV file and then using the DSL.

It runs to completion in about three minutes, so it seems to be pretty quick overall. To work "for real" it will need #173 to be closed. I should have a PR for that forthcoming here in a minute.


ahwagner:
[Finish DGIdb PRs](https://trello.com/c/iEeM6eHD/1-finish-dgidb-prs)

ahwagner:
Looks good to me. Test import ran great. +1

PR #183, opened 2017-06-08
Online Updater for DoCM

susannasiebert:


lib/genome/online_updaters/docm/api_client.rb
```diff
@@ -0,0 +1,29 @@
+module Genome; module OnlineUpdaters; module Docm
+  class ApiClient
+    def variants
+      @variants = JSON.parse(get_page(variant_url))
```
acoffman: It looks like the results here are stored in an instance variable (`@variants`) which is never referenced again. Each call to this method will make the HTTP request and just rewrite whatever was in `@variants` before. If you wanted to memoize the result, you'll want to do an `@variants ||=`.  

Though this doesn't change the correctness of the method. It will work fine as written!

lib/genome/online_updaters/docm/updater.rb
```diff
@@ -0,0 +1,132 @@
+require 'net/http'
```
acoffman: This doesn't need to be here - nothing in this module uses `net/http`. You could put it at the top of the `ApiClient` class as it actually uses `Net::HTTP` but its not technically required either place as rails already requires it.

Also this is totally my mistake, I did the same in the CIViC importer; I'll change that :)

lib/genome/online_updaters/docm/api_client.rb
```diff
@@ -0,0 +1,29 @@
...
+      'http://docm.genome.wustl.edu/api/v1/variants.json'
+    end
+
+    def get_page(url)
+      uri = URI(url)
+
+      uri.query = URI.encode_www_form(docm_params)
+
+      req = Net::HTTP::Get.new(uri)
+      resp = Net::HTTP.start(uri.host, uri.port) { |http| http.read_timeout = 1000; http.request(req)}
```
acoffman: I'm guessing you're using this form rather than `res = Net::HTTP.get_response(uri)` because the endpoint is slow and you need to specify the timeout manually?

lib/genome/online_updaters/docm/updater.rb
```diff
@@ -0,0 +1,132 @@
...
+        name: variant['gene']
+      ).first_or_create
+      #Entrez Gene Id
+    end
+
+    def parse_interaction_information(variant)
+      if !variant.key?('meta')
+        return []
+      end
+      fields = variant['meta'].first['Drug Interaction Data']['fields']
```
acoffman: For a little bit of future proofing, there's no guarantee that there's only going to be one table in the`meta` section or that the drug interaction data one will be the first one, so it might be safer to do something like:

```
return [] unless variant.include?('meta')
drug_data = variant['meta'].find { |table| table.include?('Drug Interaction Data') }
return [] unless drug_data.present?
fields = drug_data['fields']
rows = drug_data['rows']
```

lib/genome/online_updaters/docm/updater.rb
```diff
@@ -0,0 +1,132 @@
...
+        end
+        row_hash['Therapeutic Context'].split(/,|\+|plus/).each do |drug|
+          info = row_hash
+          info['Therapeutic Context'] = drug
+          if valid_drug?(drug)
+            interaction_information.add(info)
+          end
+        end
+      end
+      return interaction_information
```
acoffman: Don't need an explicit return here as its not needed for flow control. If you don't want the bare `interaction_information` on its own line, you could use a `Set.new.tap do |interaction_information|` around the `rows.each`.

lib/genome/online_updaters/docm/updater.rb
```diff
@@ -0,0 +1,132 @@
...
+      rows = variant['meta'].first['Drug Interaction Data']['rows']
+      interaction_information = Set.new()
+      rows.each do |row|
+        row_hash = {}
+        fields.zip(row).each do |name, value|
+          row_hash[name] = value
+        end
+        row_hash['Therapeutic Context'].split(/,|\+|plus/).each do |drug|
+          info = row_hash
+          info['Therapeutic Context'] = drug
```
acoffman: This works, but not how you're expecting. `info` will just point to `row_hash`, so you're just overwriting the original value each time i.e.

```
irb(main):001:0> x = { 1 => 2, 3 => 4}
=> {1=>2, 3=>4}
irb(main):002:0> y = x
=> {1=>2, 3=>4}
irb(main):003:0> y[1] = 5
=> 5
irb(main):004:0> x
=> {1=>5, 3=>4}
```

Adding it to the set does copy it rather than add a pointer:

```
irb(main):003:0> z = Set.new
=> #<Set: {}>
irb(main):004:0> x =  {1 => 2, 3 => 4}
=> {1=>2, 3=>4}
irb(main):005:0> y = x
=> {1=>2, 3=>4}
irb(main):006:0> z.add(x)
=> #<Set: {{1=>2, 3=>4}}>
irb(main):007:0> y[1] = 5
=> 5
irb(main):008:0> y
=> {1=>5, 3=>4}
irb(main):009:0> x
=> {1=>5, 3=>4}
irb(main):010:0> z.add(x)
=> #<Set: {{1=>5, 3=>4}, {1=>5, 3=>4}}>
```
So the code ends up working, but if you're just going to overwrite the original value each time, there's no need for the local `info` variable. Alternately you could could use `dup` which copies the hash which would behave the way I think you intended:

```
irb(main):001:0> x = { 1 => 2, 3 => 4}
=> {1=>2, 3=>4}
irb(main):002:0> y = x.dup
=> {1=>2, 3=>4}
irb(main):003:0> y[1] = 5
=> 5
irb(main):004:0> y
=> {1=>5, 3=>4}
irb(main):005:0> x
=> {1=>2, 3=>4}
```

lib/genome/online_updaters/docm/updater.rb
```diff
@@ -0,0 +1,132 @@
...
+    def valid_drug?(drug_name)
+      [
+        !drug_name.include?('inhib'),
+        !drug_name.include?('HER3'),
+        !drug_name.include?('TKI'),
+        !drug_name.include?('anti'),
+        !drug_name.include?('BRAF'),
+        !drug_name.include?('radio'),
+        !drug_name.include?('BH3'),
+      ].all?
```
acoffman: This totally follows the same pattern as the other importer, so its fine to leave "as is" for consistency, but you could also simplify it to something like this if you wanted:

```
['inhib', 'HER3', 'TKI', 'anti', 'BRAF', 'radio', 'BH3'].none? { |name| drug_name.include?(name) }
```

lib/genome/online_updaters/docm/updater.rb
```diff
@@ -0,0 +1,132 @@
...
+      ).first_or_create
+    end
+
+    def get_or_create_interaction_claim(gene_claim, drug_claim, interaction_info, diseases, source)
+      DataModel::InteractionClaim.where(
+        gene_claim_id: gene_claim.id,
+        drug_claim_id: drug_claim.id,
+        source_id: source.id,
+        known_action_type: 'n/a',
+      ).first_or_create.tap do |interaction_claim|
```
acoffman: The `tap` isn't strictly needed here though it will work just fine as written! `first_or_create` itself takes an optional block, and that block is only called when a new record is created so unless we expect to have duplicate interaction claims with differing attributes, we can skip the `tap`

lib/genome/online_updaters/docm/updater.rb
susannasiebert: My concern here was that if I end up overriding the actual value of `row_hash['Therapeutic Context']` while looping over the same value, that it would somehow screw up the looping. ...although it gets run through `split` so maybe I don't need to be concerned after all? I still think I would prefer to avoid overwriting the original `row_hash` entry so using `dup` would probably be the cleanest solution.

lib/genome/online_updaters/docm/api_client.rb
susannasiebert: Yeah, I got an error the first time I ran this without a timeout.

lib/genome/online_updaters/docm/updater.rb
susannasiebert: I was running into an error without the tap because `interaction_claim` didn't have an id associated with it yet, which caused the subsequent call to `interaction_claim.id` to fail when trying to create the interaction claim attribute.

susannasiebert:
@acoffman I pushed a commit to implement most of your suggested changes/fixes. See my comments above for the remaining suggestions.

ahwagner:
[Finish DGIdb PRs](https://trello.com/c/iEeM6eHD/1-finish-dgidb-prs)

PR #184, opened 2017-06-16
WIP: app/jobs for online updating of sources

susannasiebert:


app/jobs/api_updater.rb
```diff
@@ -0,0 +1,10 @@
+class ApiUpdater < Updater
+  attr_reader :updater
```
acoffman: `attr_reader` defines an accessor method for an instance variable (`@updater` in this case). If you just want to enforce that subclasses implement an `updater` method, I'd do something like 

```
def updater
  raise StandardError.new('Must implement #updater in subclass')
end
```

app/jobs/api_updater.rb
```diff
@@ -0,0 +1,10 @@
...
+
+  def perform(recurring = true)
+    begin
+      updater.update
+      #reschedule if recurring
```
acoffman: This probably doesn't need to be commented out anymore

app/jobs/tsv_updater.rb
```diff
@@ -0,0 +1,35 @@
...
+  attr_reader :importer
+
+  def perform(recurring = true)
+    begin
+      create_tempfile
+      download_file
+      importer.import
+    ensure
+      remove_download
+      #reschedule if recurring
```
acoffman: Same as above

app/jobs/tsv_updater.rb
```diff
@@ -0,0 +1,35 @@
+require 'tempfile'
+require 'open-uri'
+
+class TsvUpdater < Updater
+  attr_reader :tempfile
+  attr_reader :tempfile_name
+  attr_reader :latest_url
+  attr_reader :importer
```
acoffman: Same as above; if you just want to enforce that implementing classes define `importer` you could get rid of this and do 

```
def importer
   raise StandardError.new('Must define #importer in subclass')
end
```

or similar. Same with the other elements of this interface (tempfile name, latest url, etc).

app/jobs/updater.rb
```diff
@@ -0,0 +1,13 @@
+#class Updater  < ActiveJob::Base
+class Updater
+  def reschedule
+    self.class.set(wait_until: next_month).perform_later
+  end
+
+  def next_month
```
acoffman: As we may want to update sources with different frequencies, I might make that part of the interface that subclasses must implement.

So we could do something like:

```
def reschedule
  self.class.set(wait_until: next_update_time).perform_later
end

def next_update_time
   raise StandardError.new('must implement in subclass')
end
```

That way we could say, update civic weekly, but entrez every few months (or whatever).

susannasiebert:
@acoffman I made a few commits fixing the issues you pointed out.

PR #186, opened 2017-06-16
Rails upgrade

acoffman:
This PR takes us from rails 3.2-> rails 5. All tests are passing and as best as I can tell everything we expect to work in the interface does. It also upgrades the version of Bootstrap we're using. This has resulted in some cosmetic bugs that will require CSS tweaks. I've tried to adjust the markup to fix things as I've gone along, but there are definitely some remaining issues. 

This also incorporates @yanyangfeng's `rails4-search` branch changes as of today.


acoffman:
The tests aren't passing on travis for some reason related to loading the database schema; they're passing locally for me, so I'll look into what's different here!

app/controllers/interaction_claims_controller.rb
```diff
@@ -28,6 +27,8 @@ def interaction_search_results
         combine_input_drugs(params)
       end
     end
+    puts "YOOOOOOOOO"
```
susannasiebert: Did you mean to commit these print statements?

app/filters/filter_chain.rb
```diff
@@ -9,8 +9,13 @@ def initialize
   #TODO it will now recalculate if you modify the filter after calcuating it
   #but once something is excluded, it can't be added back in
   #is that an acceptable limitation?
+
+  # this is what defines the "include_" methods
+
   Filter.all_filters.each do |filter|
+    puts "INCLUDE EXCLUDE"
```
susannasiebert: More print statements.

app/filters/filter_chain.rb
```diff
@@ -30,9 +37,13 @@ def empty?
...
 
   def evaluate_all_filters
-    @computed_include ||= evaluate_filter(@all_include)
-    @computed_exclude ||= evaluate_filter(@all_exclude)
-    @computed_final ||= @computed_include - @computed_exclude
+    #@computed_include ||= evaluate_filter(@all_include)
+    #@computed_exclude ||= evaluate_filter(@all_exclude)
+    #@computed_final ||= @computed_include - @computed_exclude
+    @computed_final ||= evaluate_filter(@all_include)
+    puts "NOOOOOOOO"
```
susannasiebert: More print statements.

app/searches/lookup_categories.rb
```diff
@@ -16,8 +16,9 @@ def self.gene_names_in_category(category_name)
     DataModel::GeneClaimCategory
```
susannasiebert: Probably outside of the scope of this PR but this subroutine should retrieve gene categories directly.

susannasiebert:
One of the first failures from Travis is a failure to run `db:structure:load`. Makes me wonder if the data dump this PR is using is outdated?

susannasiebert:
I had another thought for the travis failures: do you need to update the `.travis.yml` file to incorporate the new versions?

acoffman:
@susannasiebert the tests are now passing and i've removed the erroneously committed print statements!

ahwagner:
I get a `unrecognized configuration parameter "idle_in_transaction_session_timeout"` error when I try to load this in postgres 9.4. This is because it's introduced in 9.6. @acoffman mentioned that we don't have >=9.6 on the production server, so this needs to be addressed.

ahwagner:
well, using an updated version of postgres (which we should also do on our production server), this does load, and work (mostly). Some issues are apparent now, but this is good enough to merge into master and create fixes around them.

PR #188, opened 2017-06-21
Add post-import processing to auto-import workflow

### LINKED ISSUE #178: Consider implementing a post-grouping cleanup step
We should discuss adding a clean-up step after running groupers to delete orphaned claims (claims that could not get grouped). Orphaned claims increase the runtime of the groupers which will impact the automated updating process.

This is related to #148

susannasiebert:
Closes #178

### DIFF
--- a/app/jobs/api_updater.rb
+++ b/app/jobs/api_updater.rb
@@ -3,6 +3,10 @@ class ApiUpdater < Updater
     begin
       updater.update
       reschedule if recurring
+      grouper = Grouper.new()
+      grouper.perform(group_genes = should_group_genes?, group_drugs = should_group_drugs?)
+      post_grouper = PostGrouper.new()
+      post_grouper.perform(cleanup_gene_claims = should_cleanup_gene_claims?)
     end
   end
 
--- a/app/jobs/api_updater.rb
+++ b/app/jobs/api_updater.rb
@@ -4,9 +4,9 @@ class ApiUpdater < Updater
       updater.update
       reschedule if recurring
       grouper = Grouper.new()
-      grouper.perform(group_genes = should_group_genes?, group_drugs = should_group_drugs?)
+      grouper.perform(should_group_genes?, should_group_drugs?)
       post_grouper = PostGrouper.new()
-      post_grouper.perform(cleanup_gene_claims = should_cleanup_gene_claims?)
+      post_grouper.perform(should_cleanup_gene_claims?)
     end
   end
 
--- a/app/jobs/post_grouper.rb
+++ b/app/jobs/post_grouper.rb
@@ -10,7 +10,7 @@ class PostGrouper
   def delete_orphaned_gene_claims
       DataModel::GeneClaim.joins(:gene, :gene_claim_aliases, :source)
           .where("gene_claims.gene_id IS NULL and sources.source_db_name = 'Ensembl'")
-          .each {|c| c.detroy}
+          .destroy_all
   end
 
   def update_counts
--- a/app/jobs/post_grouper.rb
+++ b/app/jobs/post_grouper.rb
@@ -1,0 +1,19 @@
+#class PostGrouper < ActiveJob::Base
+class PostGrouper
+  def perform(cleanup_gene_claims = true)
+    if cleanup_gene_claims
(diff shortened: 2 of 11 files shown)


app/jobs/update_civic.rb
```diff
@@ -2,4 +2,23 @@ class UpdateCivic < ApiUpdater
...
+
+  def next_update_time
+    Date.today
+      .beginning_of_week
+      .next_day(14)
+      .midnight
+  end
+
+  def should_group_genes?
+    return true
```
acoffman: This will obviously work just fine and there's no need to change it, but idiomatically, `return` should really only be used for flow control in ruby. The method will return the return value of the last line anyways.

app/jobs/post_grouper.rb
```diff
@@ -0,0 +1,19 @@
...
+  def perform(cleanup_gene_claims = true)
+    if cleanup_gene_claims
+      delete_orphaned_gene_claims()
+    end
+    update_counts()
+  end
+
+  def delete_orphaned_gene_claims
+      DataModel::GeneClaim.joins(:gene, :gene_claim_aliases, :source)
+          .where("gene_claims.gene_id IS NULL and sources.source_db_name = 'Ensembl'")
```
acoffman: Instead of looping over each claim and calling `destroy` which will generate `n` `delete from blah where id = ?` SQL queries, you could just do `Model.where().destroy_all` which will trigger a single `delete from blah where id in (list);`

app/jobs/api_updater.rb
```diff
@@ -3,6 +3,10 @@ def perform(recurring = true)
     begin
       updater.update
       reschedule if recurring
+      grouper = Grouper.new()
+      grouper.perform(group_genes = should_group_genes?, group_drugs = should_group_drugs?)
```
acoffman: I'm not sure this is doing what you expect - you're declaring a local variable called `group_genes` and setting it to the return value of `should_group_genes?` and then passing that value in as the first positional argument. It would be equivalent to just passing `should_group_genes?` If you were looking to use named arguments, the syntax is slightly different (and I think, unfortunately, incompatible with the version of ruby running on the production server).

app/jobs/tsv_updater.rb
```diff
@@ -9,6 +9,10 @@ def perform(recurring = true)
       create_tempfile
       download_file
       importer.import
+      grouper = Grouper.new()
+      grouper.perform(group_genes = should_group_genes?, group_drugs = should_group_drugs?)
```
acoffman: Same comment about the named arguments from above

susannasiebert:
@acoffman I believe I addressed your comments with the commits I just pushed.

app/jobs/post_grouper.rb
```diff
@@ -0,0 +1,19 @@
+#class PostGrouper < ActiveJob::Base
+class PostGrouper
+  def perform(cleanup_gene_claims = true)
```
ahwagner: default here should be false. We shouldn't be deleting gene or drug claims unless it's a special circumstance (e.g. _ensembl_).

susannasiebert: Fixed. The delete_orphaned_gene_claims will only delete gene claims from Ensembl so that's an additional safeguard. In your opinion, is the `should_cleanup_gene_claims?` `true` in `app/jobs/update_entrez.rb` correct?

ahwagner: Yeah, I saw that, and didn't think that it mattered. Since Entrez doesn't even have gene claims, it's kind of moot, right?

susannasiebert: No, because `delete_orphaned_gene_claims` doesn't delete claims from all sources, or only the currently imported source; it specifically deletes ungrouped claims from Ensembl. So in this case the Entrez import happens, grouping happens, and the during post-grouping all claims from Ensembl that we weren't able to group are getting deleted.

susannasiebert: For clarity I can rename `delete_orphaned_gene_claims` to `delete_orphaned_ensembl_gene_claims` and `should_cleanup_gene_claims?` to `should_cleanup_ensembl_gene_claims?`.

PR #190, opened 2017-06-23
First stab at cleaning up the interaction search

susannasiebert:
The main search now takes ~3 seconds instead of ~15.


app/searches/lookup_drugs.rb
```diff
@@ -29,8 +30,9 @@ def self.match_search_terms_to_objects(search_terms, scope)
     search_terms = search_terms.dup
     drug_results = DataModel::Drug.send(scope).where(name: search_terms)
     search_terms = search_terms - drug_results.map(&:name)
-    drug_alias_results = DataModel::DrugClaimAlias.send(scope).where(alias: search_terms)
+    drug_alias_results = DataModel::DrugAlias.send(scope).where(alias: search_terms)
     search_terms = search_terms - drug_alias_results.map(&:alias)
+    #Can we get rid of this case
```
ahwagner: I don't think so? We've associated all of the drug claim info we're interested in searching to the drug level--we probably don't need `for_search` in `drug_claim.rb` at all anymore, although I'd be interested to hear @acoffman and @yanyangfeng's thoughts on this.

app/searches/lookup_related_drugs.rb
```diff
@@ -3,12 +3,11 @@ class LookupRelatedDrugs
   def self.find(drug_name)
     drug_name = drug_name.gsub('(', '\(').gsub(')', '\)').gsub(':', '\:')
     drugs = DataModel::Drug.advanced_search(name: drug_name)
+    #Can we get rid of thise case?
```
ahwagner: Same as previous comment re: GeneClaim. I think it's good for us to move away from searching against claims. Ideally, we want them to appear only in the full interactions.tsv dump for now.

spec/searches/lookup_related_drugs_spec.rb
```diff
@@ -33,11 +33,10 @@
     expect(results.first.instance_variable_get("@drug")).to eq(drug)
   end
 
-  it 'should find drugs that have drug claim aliases with matching names' do
+  it 'should find drugs that have drug aliases with matching names' do
     Fabricate(:drug, name: 'turkey')
     drug = Fabricate(:drug, name: 'sandwich meats')
-    drug_claim = Fabricate(:drug_claim, drug: drug, name: 'lunchable')
-    Fabricate(:drug_claim_alias, drug_claim: drug_claim, alias: 'turkey sandwich')
+    Fabricate(:drug_alias, drug: drug, alias: 'turkey sandwich')
```
ahwagner: 🤣

app/searches/lookup_drugs.rb
susannasiebert: I'm confused. You say "I don't think so?" but the rest of your comment sounds like we can indeed get rid of this case to search against the claim.

ahwagner: It is my opinion that we can drop this, but I left the '?' because I'd like the others to weigh in on this first. I can't think of a reason we would want to keep this, but @acoffman wrote most of the search engine, and then I used that as a template for adding drug search.

obigriffith:
This was substantially improved and pagination in a separate issue will also improve. Closing for now. Re-open if people still find too slow.

PR #191, opened 2017-06-26
Chembl drug grouper

ahwagner:
Still in progress, but chembl import and grouper logic exist and first major tests pass. Additional tests to be written still.


db/migrate/20170605204001_create_chembl_molecules.rb
```diff
@@ -0,0 +1,40 @@
+class CreateChemblMolecules < ActiveRecord::Migration
+  def change
+    create_table :chembl_molecules do |t|
+      t.integer :molregno
+      t.string :pref_name, :limit => 255
```
acoffman: Its totally your call (and I wouldn't probably bother to go back and change this at this point), but I generally would lean towards using `text` rather than `string` (which will end up as `varchar`under the hood) in the future. There's no performance disadvantage to using `text` in posgres and it doesn't require you to specify a limit in advance (see the "tip" box here: https://www.postgresql.org/docs/current/static/datatype-character.html).

lib/genome/groupers/drug_grouper.rb
```diff
@@ -1,154 +1,149 @@
...
-          puts 'add members'
-          add_members
-          puts 'add drug aliases'
-          add_aliases
-          puts 'add drug attributes'
-          add_attributes
+      def run
+        begin
+          newly_added_claims_count = 0
+          drug_claims_not_in_groups.in_groups_of(1000, fill_with=false) do |claims|
```
acoffman: This is syntactically valid, but not doing what I think you intend. If you mean to use a named parameter, the syntax is different; this is simply declaring a variable named `fill_with`, setting it to `false` and then passing that in as the second positional argument.

lib/genome/groupers/drug_grouper.rb
```diff
@@ -1,154 +1,149 @@
...
+            ActiveRecord::Base.transaction do
+              grouped_claims = add_members(claims)
+              newly_added_claims_count += grouped_claims.length
+            end
+          end
+        end until newly_added_claims_count == 0
```
acoffman: I don't think I've ever seen a bottom checking loop used in ruby hah! But it works well here for sure.

lib/genome/groupers/drug_grouper.rb
```diff
@@ -1,154 +1,149 @@
...
+      end
+
+      def indirect_multimatch
+        @indirect_multimatch ||= Set.new
+      end
+
+      def drug_claims_not_in_groups
+        DataModel::DrugClaim.where(drug_id: nil).to_a.keep_if do |drug_claim|
+          !(
+            direct_multimatch.member? drug_claim or
```
acoffman: Generally speaking, you'll want to use `||` rather than `or` here. They are similar, but not actually equivalent in ruby. The english words bind at a lower precedence and are more appropriately used for say, flow control `do_something or raise Exception` and the symbolic ones are more appropriate for boolean logic like this.

lib/genome/updaters/chembl.rb
```diff
@@ -0,0 +1,73 @@
...
+      def self.connection
+        @@connection ||= PG.connect(dbname: chembl_db_name)
+      end
+
+      def self.init_records_from_db # This is for an initial, non-incremental load into the chembl_molecule table
+        new_records = []
+        i = 0
+        connection.exec( "SELECT * FROM molecule_dictionary").each do |record|
+          new_records << get_params(record, mol_col_hash)
+          unless i % 50000
```
acoffman: Everything but `nil` and `false` is `true` in ruby, so unless I'm missing something here; this will be true every single time it runs, regardless of the value of `i`.

lib/utils/database.rb
```diff
@@ -196,5 +222,15 @@ def self.destroy_na
 
       ActiveRecord::Base.connection.execute(sql)
     end
+
+    def self.model_to_tsv model
```
acoffman: This is a nifty little function!

lib/genome/groupers/drug_grouper.rb
ahwagner: I do think that this is doing what I expect (based on documentation from https://apidock.com/rails/Array/in_groups_of), although I agree that at first glance it seems wrong, and should instead by something like `fill_with: false`. Sample IRB below. 

```ruby
[6] pry(main)> >> a = [1, 3, 5, 7, 9]
=> [1, 3, 5, 7, 9]
[7] pry(main)> >> a.in_groups_of(2)
=> [[1, 3], [5, 7], [9, nil]]
[8] pry(main)> >> a.in_groups_of(2, fill_with=false)
=> [[1, 3], [5, 7], [9]]
[9] pry(main)> >> a.in_groups_of(2, fill_with: false)
=> [[1, 3], [5, 7], [9, {:fill_with=>false}]]
[10] pry(main)> >> a.in_groups_of(2, false)
=> [[1, 3], [5, 7], [9]]
```

db/migrate/20170605204001_create_chembl_molecules.rb
ahwagner: That's neat, I didn't know that. I made these specifications based on the existing specs from the ChEMBLdb, but if I ever have to adjust these (or create new such fields in the future), I'll just stick with `text`. Thanks!

lib/genome/groupers/drug_grouper.rb
ahwagner: Yeah, @susannasiebert clued me into this one. It does fit the bill!

lib/genome/groupers/drug_grouper.rb
acoffman: Oh yeah, sorry if that was unclear! Its functionally working fine, it just isn't using a named parameter as I assumed you were intending. Its just using the last form of your examples here and passing in `false` as the second positional argument.

ahwagner: Ah, I see. I'm actually assigning a variable `fill_with` to value `false`, and that's getting passed in.

I understand what you were saying now. Good catch, definitely not my intent.

lib/utils/database.rb
ahwagner: Yeah, and I don't think I use it anywhere else in the codebase 🤦‍♂️ .

Oh, well, I'm sure I had good reason at the time?

acoffman:
Looks good!

spec/lib/groupers/drug_grouper_spec.rb
```diff
@@ -148,11 +179,50 @@ def create_chembl_drug_attributes
...
+    source = Fabricate(:source, source_db_name: 'ChEMBL')
+    chembl_drug_claim = Fabricate(:drug_claim, source: source)
+    chembl_drug = Fabricate(:drug, name: chembl_drug_claim.primary_name)
+
+    grouper = Genome::Groupers::DrugGrouper.new
+    grouper.run
+    anti_neoplastic_drugs.each do |drug|
+      drug.reload
+      expect(drug.drug_claims.count).to eq 1
+      drug.send(:update_anti_neoplastic_add, drug.drug_claims.first) # Why does this need to be done for rspec?
```
ahwagner: There's something not quite right here. In a Rails console session, I can get the `Drug` `after_add` callback to work properly when adding a drug claim to a drug, but in rspec I need to call the associated private method manually to do the update.

```ruby
Loading development environment (Rails 5.1.1)
[1] pry(main)> >> drug = DataModel::Drug.new
=> #<DataModel::Drug:0x007f9264832658
 id: nil,
 name: nil,
 fda_approved: nil,
 immunotherapy: nil,
 anti_neoplastic: nil,
 chembl_id: nil>
[2] pry(main)> >> generic_source = DataModel::Source.where(source_db_name: 'ChEMBL').first
<Model::Source.where(source_db_name: 'ChEMBL').first
  DataModel::Source Load (0.9ms)  SELECT  "sources".* FROM "sources" WHERE "sources"."source_db_name" = $1 ORDER BY "sources"."id" ASC LIMIT $2  [["source_db_name", "ChEMBL"], ["LIMIT", 1]]
=> #<DataModel::Source:0x007f925d50fb80
 id: "30409f4e-3eb6-46fd-9661-9e8bfd6f525f",
 source_db_name: "ChEMBL",
 source_db_version: "chembl_20",
 citation:
  "The ChEMBL bioactivity database: an update. Bento AP, Gaulton A, Hersey A, Bellis LJ, Chambers J, Davies M, Krüger FA, Light Y, Mak L, McGlinchey S, Nowotka M, Papadatos G, Santos R, Overington JP. Nucleic Acids Res. 42(Database issue):D1083-90. PubMed ID: 24214965",
 base_url: "https://www.ebi.ac.uk/chembldb/index.php/target/inspect/",
 site_url: "https://www.ebi.ac.uk/chembl",
 full_name: "The ChEMBL Bioactivity Database",
 source_type_id: "ce28974a-b813-4717-a3d2-2df0cc54ab5a",
 gene_claims_count: 268,
 drug_claims_count: 906,
 interaction_claims_count: 1192,
 interaction_claims_in_groups_count: 0,
 gene_claims_in_groups_count: 252,
 drug_claims_in_groups_count: 748,
 source_trust_level_id: "600d121e-6260-4db8-9689-e72a13cbf2c3",
 gene_gene_interaction_claims_count: 0>
[3] pry(main)> >> cancer_source = DataModel::Source.where(source_db_name: 'TALC').first
<taModel::Source.where(source_db_name: 'TALC').first
  DataModel::Source Load (1.2ms)  SELECT  "sources".* FROM "sources" WHERE "sources"."source_db_name" = $1 ORDER BY "sources"."id" ASC LIMIT $2  [["source_db_name", "TALC"], ["LIMIT", 1]]
=> #<DataModel::Source:0x007f925f6b19c8
 id: "0009971b-b332-44a2-82a7-45128289099d",
 source_db_name: "TALC",
 source_db_version: "12-May-2016",
 citation:
  "A tabulated summary of targeted and biologic therapies for non-small-cell lung cancer. Simon GR, Somaiah N. Clin Lung Cancer. 2014 Jan;15(1):21-51. doi: 10.1016/j.cllc.2013.11.009. Epub 2013 Nov 21. PMID: 24377743",
 base_url: "http://www.ncbi.nlm.nih.gov/gene?term=",
 site_url: "https://www.ncbi.nlm.nih.gov/pubmed/24377743",
 full_name: "Targeted Agents in Lung Cancer (Commentary, 2014)",
 source_type_id: "ce28974a-b813-4717-a3d2-2df0cc54ab5a",
 gene_claims_count: 0,
 drug_claims_count: 0,
 interaction_claims_count: 0,
 interaction_claims_in_groups_count: 0,
 gene_claims_in_groups_count: 0,
 drug_claims_in_groups_count: 0,
 source_trust_level_id: "600d121e-6260-4db8-9689-e72a13cbf2c3",
 gene_gene_interaction_claims_count: 0>
[4] pry(main)> >> dc = DataModel::DrugClaim.new(source: generic_source)
=> #<DataModel::DrugClaim:0x007f925fde9588
 id: nil,
 name: nil,
 nomenclature: nil,
 source_id: "30409f4e-3eb6-46fd-9661-9e8bfd6f525f",
 primary_name: nil,
 drug_id: nil>
[5] pry(main)> >> cancer_dc = DataModel::DrugClaim.new(source: cancer_source)
=> #<DataModel::DrugClaim:0x007f925fdc65d8
 id: nil,
 name: nil,
 nomenclature: nil,
 source_id: "0009971b-b332-44a2-82a7-45128289099d",
 primary_name: nil,
 drug_id: nil>
[6] pry(main)> >> drug.drug_claims << dc
=> [#<DataModel::DrugClaim:0x007f925fde9588
  id: nil,
  name: nil,
  nomenclature: nil,
  source_id: "30409f4e-3eb6-46fd-9661-9e8bfd6f525f",
  primary_name: nil,
  drug_id: nil>]
[8] pry(main)> >> drug.anti_neoplastic
=> false
[9] pry(main)> >> drug.drug_claims << cancer_dc
=> [#<DataModel::DrugClaim:0x007f925fde9588
  id: nil,
  name: nil,
  nomenclature: nil,
  source_id: "30409f4e-3eb6-46fd-9661-9e8bfd6f525f",
  primary_name: nil,
  drug_id: nil>,
 #<DataModel::DrugClaim:0x007f925fdc65d8
  id: nil,
  name: nil,
  nomenclature: nil,
  source_id: "0009971b-b332-44a2-82a7-45128289099d",
  primary_name: nil,
  drug_id: nil>]
[10] pry(main)> >> drug.anti_neoplastic
=> true
[11] pry(main)> >> drug.drug_claims.last.drug_id = nil
=> nil
[12] pry(main)> >> drug.anti_neoplastic
=> true
[13] pry(main)> >> drug.drug_claims
=> [#<DataModel::DrugClaim:0x007f925fde9588
  id: nil,
  name: nil,
  nomenclature: nil,
  source_id: "30409f4e-3eb6-46fd-9661-9e8bfd6f525f",
  primary_name: nil,
  drug_id: nil>,
 #<DataModel::DrugClaim:0x007f925fdc65d8
  id: nil,
  name: nil,
  nomenclature: nil,
  source_id: "0009971b-b332-44a2-82a7-45128289099d",
  primary_name: nil,
  drug_id: nil>]
```
Also, I noticed that for _removing_ drug claims from a drug, the `before_remove` callback also doesn't seem to work, even from console, unless the private method is called directly from the object:
```ruby
[14] pry(main)> >> drug.drug_claims.delete(cancer_dc)
=> [#<DataModel::DrugClaim:0x007f925fdc65d8
  id: nil,
  name: nil,
  nomenclature: nil,
  source_id: "0009971b-b332-44a2-82a7-45128289099d",
  primary_name: nil,
  drug_id: nil>]
[15] pry(main)> >> drug.anti_neoplastic
=> true
[17] pry(main)> >> drug.send(:update_anti_neoplastic_remove, 'foobar')
=> false
[18] pry(main)> >> drug.anti_neoplastic
=> false
```

@acoffman, could you shed some light on what's going on here? I don't think that this is a show-stopper for our current work, but it would be nice for my understanding as to why these specific callbacks seem to be finicky.

ahwagner:
This PR is ready for final review, @acoffman and @susannasiebert.

ahwagner:
Hey all. A few more commits went in--minor stuff, things that I missed in my initial testing that broke grouping on real data. Grouper has been running steady for about 10 minutes now, so I think that a final review at the hackathon is in order.

ahwagner:
Well, about 1/3 of the drug claims had been grouped before there was an unexpected failure (relating to an 'α' character in an alias). I've committed the grouped claims, and am running the interaction grouper now. This is not a publishable state of the database, but it should give us something to work with for testing. I'm running the interaction grouper now.

susannasiebert:
I think it looks good.

PR #194, opened 2017-06-29
WIP: Drugbank update

kcotto:
Importer modified by commenting out the `def self.get_version` in order to import current drugbank contents into db before merge. Code will be reverted back once versioning has been fixed in the respective scripts (which I'm working on but just haven't got working up until this point).


kcotto:
Importer reverted back to original after python scripts were updated and working.

PR #204, opened 2017-06-30
Fix drug_grouper ChemblMoleculeSynonyms alias creation

susannasiebert:


susannasiebert:
Should we add a test that catches the first_or_create error? Not sure what that would look like.

ahwagner:
Yeah, that could be worthwhile, and not hard to make. On it.

PR #213, opened 2017-07-05
TSV exports

ahwagner:
Our old method of exporting a TSV using jquery datatables no longer works with our 3.0 views. This PR creates a new method for constructing a TSV string from the result data, and leverages the existing utility for generating and returning the file from the string.

This PR is not ready for merging. It requires some optimization (interaction source roll-up and eager loading) of the objects returned by the interaction search. However, it would be good to do a quick review of the existing code to ensure that nothing dangerous/ill-conceived is going on here.

### DIFF
--- a/app/controllers/interaction_claims_controller.rb
+++ b/app/controllers/interaction_claims_controller.rb
@@ -79,9 +79,9 @@ class InteractionClaimsController < ApplicationController
           row_hash['match_term'] = identifiers.select { |i| i.id == id}.first.name
           row_hash['gene'] = interaction.gene.name
           row_hash['drug'] = interaction.drug.name
-          row_hash['interaction_types'] = interaction.interaction_types.pluck(:type).uniq.join(compound_field_separator)
-          row_hash['pmids'] = interaction.publications.pluck(:pmid).uniq.join(compound_field_separator)
-          row_hash['sources'] = interaction.sources.pluck(:source_db_name).uniq.join(compound_field_separator)
+          row_hash['interaction_types'] = interaction.type_names.join(compound_field_separator)
+          row_hash['pmids'] = interaction.pmids.join(compound_field_separator)
+          row_hash['sources'] = interaction.source_names.join(compound_field_separator)
           tsv << headers.map{ |field| row_hash[field] }
         end
       end
--- a/app/controllers/interaction_claims_controller.rb
+++ b/app/controllers/interaction_claims_controller.rb
@@ -71,17 +71,18 @@ class InteractionClaimsController < ApplicationController
             'match_type' => result.match_type_label,
             'search_term' => result.instance_variable_get(:@search_term)
         }
-        result.instance_variable_get(:@identifiers).each do |id|
-          row_hash['match_term'] = id.name
-          result.instance_variable_get(:@interactions).each do |interaction|
-            next unless id.id == interaction[match_field]
-            row_hash['gene'] = interaction.gene.name
-            row_hash['drug'] = interaction.drug.name
-            row_hash['interaction_types'] = interaction.interaction_types.pluck(:type).uniq.join(compound_field_separator)
-            row_hash['pmids'] = interaction.publications.pluck(:pmid).uniq.join(compound_field_separator)
-            row_hash['sources'] = interaction.interaction_claims.map{ |ic| ic.source.source_db_name }.uniq.join(compound_field_separator)
-            tsv << headers.map{ |field| row_hash[field] }
-          end
+        identifiers = result.instance_variable_get(:@identifiers)
+        # result.instance_variable_get(:@identifiers).each do |id|
+        #   row_hash['match_term'] = id.name
+        result.instance_variable_get(:@interactions).each do |interaction|
+          id = interaction[match_field]
+          row_hash['match_term'] = identifiers.select { |i| i.id == id}.first.name
+          row_hash['gene'] = interaction.gene.name
+          row_hash['drug'] = interaction.drug.name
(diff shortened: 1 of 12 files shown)


app/controllers/interaction_claims_controller.rb
```diff
@@ -55,4 +51,39 @@ def perform_interaction_search
...
+            'match_type' => result.match_type_label,
+            'search_term' => result.instance_variable_get(:@search_term)
+        }
+        result.instance_variable_get(:@identifiers).each do |id|
+          row_hash['match_term'] = id.name
+          result.instance_variable_get(:@interactions).each do |interaction|
+            next unless id.id == interaction[match_field]
+            row_hash['gene'] = interaction.gene.name
+            row_hash['drug'] = interaction.drug.name
+            row_hash['interaction_types'] = interaction.interaction_types.pluck(:type).uniq.join(compound_field_separator)
```
susannasiebert: The implementation to get the interaction_types, pmids, and sources from interactions might be something to useful to include on interactions directly.

ahwagner:
Okay, this is ready for a full review @susannasiebert and @acoffman. As a bonus, overall search speed has been improved.

app/controllers/interaction_claims_controller.rb
ahwagner: Ah, I think I misunderstood your comment here. You mean write these methods directly into the interaction model? i.e. `interaction.type_names`, `interaction.source_names`, `interaction.pmids`? I like that idea, but I want to make sure that's what you mean.

susannasiebert: Yes, that's what I meant. Sorry for the confusion. I expect these to be needed in several other places as well so putting them in the model directly should reduce duplication.

app/controllers/interaction_claims_controller.rb
```diff
@@ -55,4 +51,40 @@ def perform_interaction_search
     search_results = LookupInteractions.find(params)
     @search_results = InteractionSearchResultsPresenter.new(search_results, view_context)
   end
+
+  def unpack_locals params
```
acoffman: Probably want to add `()` around params here. Its syntactically valid, but by convention, parens are always used in method declarations, and optional when calling methods.

app/controllers/interaction_claims_controller.rb
```diff
@@ -55,4 +51,40 @@ def perform_interaction_search
...
+    @preset_clin = (params[:clinically_actionable] == "checked" ? "Clinically Actionable" : "")
+    @preset_druggable = (params[:druggable_genome] == "checked" ? "Druggable Genome" : "")
+    @preset_resist = (params[:drug_resistance] == "checked" ? "Drug Resistance" : "")
+  end
+
+  def prepare_export(compound_field_separator = '|')
+    headers = %w[search_term match_term match_type gene drug interaction_types sources pmids]
+    @tsv = CSV.generate(col_sep: "\t") do |tsv|
+      tsv << headers
+      @search_results.instance_variable_get(:@search_results).each do |result|
```
acoffman: Rather than using reflection here with `instance_variable_get` you could always just add an accessor method in the search results class `attr_reader :search_results` and then call `@search_results.search_results` directly. Likewise for `type`, `search_term`, etc

app/controllers/interaction_claims_controller.rb
```diff
@@ -55,4 +51,40 @@ def perform_interaction_search
...
+        match_field = result.instance_variable_get(:@type).singularize + '_id'
+        row_hash = {
+            'match_type' => result.match_type_label,
+            'search_term' => result.instance_variable_get(:@search_term)
+        }
+        identifiers = result.instance_variable_get(:@identifiers)
+        # result.instance_variable_get(:@identifiers).each do |id|
+        #   row_hash['match_term'] = id.name
+        result.instance_variable_get(:@interactions).each do |interaction|
```
acoffman: This is probably fine for now, but in general, this is logic that maybe belongs in a view or a presenter as it deals with representing a model object in some view-able format. DoCM has some decently clear TSV exporters you could look to for inspiration:

Controller: https://github.com/genome/docm/blob/master/app/controllers/api_v1_controller.rb
Corresponding view: https://github.com/genome/docm/blob/master/app/views/api_v1/variants.tsv.erb

In the same repo, the ClinVar exporter does a similar thing with even more complicated logic (including a presenter and customized filename for the tsv) if you wanted to take a look at that too.

app/models/data_model/drug.rb
```diff
@@ -15,7 +15,8 @@ class Drug < ::ActiveRecord::Base
 
 
     def self.for_search
-      eager_load(drug_claims: {interaction_claims: { source: [], gene_claim: [:source, :gene_claim_categories], interaction_claim_types: [], drug_claim: [drug: [drug_claims: [:drug_claim_types]]]}})
+      # eager_load(drug_claims: {interaction_claims: { source: [], gene_claim: [:source, :gene_claim_categories], interaction_claim_types: [], drug_claim: [drug: [drug_claims: [:drug_claim_types]]]}})
```
acoffman: Once you're comfortable that these are working as intended, I'd removed the commented out lines (likewise for all the commented lines in this PR).

app/controllers/interaction_claims_controller.rb
ahwagner: Yeah, I agree with what you're saying--the controller felt like a weird place to put this. Is there a way that I could pull this logic out to a presenter, but retain the data I need (the search results) without performing a second search?

ahwagner:
I'm moving forward with the merge and moving the discussion of https://github.com/griffithlab/dgi-db/pull/213#discussion_r125949641 into #215.

Issue #92, opened 2015-11-06
Fix interaction claim type view


ahwagner:
On detailed results page, remove links from "interaction type" and make source links go to interaction claim.

On Summary results page, change checkboxes to interaction symbols that link to interaction claims.

ahwagner:
this will be addressed as part of #149

ahwagner:
#165 breaks links out to interaction claims from the detailed results view in interaction search results.

ahwagner:
add 'n/a' as a default entry for interaction type on results display.

yanyangfeng:
This should be fixed

Issue #88, opened 2015-11-06
Revise gene and drug claim views to appear more like gene and drug views


malachig:
e.g.
http://dgidb.genome.wustl.edu/drug_claims/GuideToPharmacologyInteractions/4941
http://dgidb.genome.wustl.edu/gene_claims/GuideToPharmacologyInteractions/1797

ahwagner:
This is going to be revised as part of @yanyangfeng's updates in #124

yanyangfeng:
I think these are looking better on my dev instance. What do people think, can we close this issue?

<img width="1351" alt="screen shot 2017-05-05 at 4 07 15 pm" src="https://cloud.githubusercontent.com/assets/16638655/25764355/0201dce4-31ad-11e7-9a02-a1320383548a.png">
<img width="1344" alt="screen shot 2017-05-05 at 4 06 49 pm" src="https://cloud.githubusercontent.com/assets/16638655/25764358/034f7eee-31ad-11e7-8804-4ac3bce98202.png">

ahwagner:
I think that we should refactor these views to look more like the new drug and gene group views. I'll revise the issue accordingly.

yanyangfeng:
I believe this can be closed

Issue #67, opened 2015-10-06
Keep previously viewed tab on results page in foreground when returning to page


ahwagner:
When you press the browser "back" button from a result, it should return you to the previous tab (such as "primary interactions"). Instead, the default view ("by source") is displayed.

ahwagner:
This is partially addressed in #103, although it only takes you back to the previous tab, not necessarily the same page of results on that tab.

ahwagner:
part of #124

yanyangfeng:
I believe this is done.

PR #221, opened 2017-07-18
Rails4 search - drug gene claim views and na default interaction type

yanyangfeng:
Addresses issues #92 and #88


yanyangfeng:
Pushed changes to this branch to address issue #67 as well

Issue #219, opened 2017-07-07
update ChEMBL interactions


ahwagner:
ChEMBL drugs are good, but the interactions need to be updated from ChEMBL v20.

ahwagner:
closed in #227

PR #227, opened 2017-07-27
Chembl update

ahwagner:
This is the updater for ChEMBL interactions.


ahwagner:
The Chembl Updater claims grouped well (which is to be expected). All of the drug claims grouped (yay!) and all but 2 gene claims grouped (due to multi-alias match). All but 4 of the interaction claims grouped.

ahwagner:
@yanyangfeng good catch. An association exists (https://github.com/griffithlab/dgi-db/blob/5f55aab0560b85f7a327a661f9aaf6635b26dc50/app/models/data_model/drug.rb#L10), but no FK is present in the database. I'll make a migration to add that.

ahwagner:
closes #219

ahwagner:
@yanyangfeng there are some fields (but not all) that we could add unique indices to, I'll make those changes as well to tighten up the tables a bit.

PR #228, opened 2017-07-28
Drug bank online updater

susannasiebert:
There is a small bug with the drug grouper that makes the post-processing fail for some drug attributes with special characters but aside from that, this online updater should work as expected.


ahwagner:
It looks like you will need to add your Gemfile.lock to this PR to get tests to pass.

lib/genome/importers/drug_bank/new_drug_bank.rb
```diff
@@ -0,0 +1,127 @@
...
+    def create_source
+      @source = DataModel::Source.where(
+        base_url: 'http://www.drugbank.ca',
+        site_url: 'http://www.drugbank.ca/',
+        citation: "DrugBank 4.0: shedding new light on drug metabolism. Law V, Knox C, Djoumbou Y, Jewison T, Guo AC, Liu Y, Maciejewski A, Arndt D, Wilson M, Neveu V, Tang A, Gabriel G, Ly C, Adamjee S, Dame ZT, Han B, Zhou Y, Wishart DS.Nucleic Acids Res. 2014 Jan 1;42(1):D1091-7. PubMed ID: 24203711",
+        source_db_version: get_version,
+        source_type_id: DataModel::SourceType.INTERACTION,
+        source_db_name: 'DrugBank',
+        full_name: 'DrugBank - Open Data Drug & Drug Target Database'
+        #source_trust_level_id: DataModel::SourceTrustLevel.EXPERT_CURATED
```
kcotto: Should we comment this code out and change the source level to non-curated since we are moving to online updating? Or is it easier to just change that within the database itself?

ahwagner: This commented line could be removed entirely. It's worth noting that we're not really tracking / using that field anymore, and at some point, we'll remove this field entirely.

It appears that we currently list DrugBank as "non-curated" [here](https://github.com/griffithlab/dgi-db/blob/master/lib/genome/normalizers/source_trust_level.rb#L33).

kcotto:
I will get that updated today.

lib/genome/importers/drug_bank/new_drug_bank.rb
```diff
@@ -0,0 +1,127 @@
...
+      ).first_or_create
+    end
+
+    def import_claims
+      blank_filter = ->(x) { x.blank? || x == "''" || x == '""' || x.downcase.tr('/','') == 'na'}
```
ahwagner: @susannasiebert Do lines 43-45 need to be here? They look like an artifact from the old tsv grouper.

ahwagner: I removed these lines in https://github.com/griffithlab/dgi-db/pull/228/commits/532258c43887b6c1b18f0daef7bb2c45d9be09b2 and it didn't seem to affect performance. If they're needed for some reason, feel free to revert.

ahwagner:
Alternatively (and this is my preferred option), we set the default value to false in `updater.rb`: https://github.com/susannasiebert/dgi-db/blob/drug_bank/app/jobs/updater.rb#L20

susannasiebert:
I believe this PR is ready to get merged in.

yanyangfeng:
this looks good to me

Issue #218, opened 2017-07-07
Create .tsv Downloads of Drugs and Genes


need47:
First of all, thanks for your hard work on this comprehensive resource. I can see a lot of things going on here. 

It seems that you guys have decided to use ChEMBL id for grouping drugs (based on drugs.chembl_id). That's a good start. I used to check a type field for "PubChem CID" in drug_aliases table in order to get the CID. Now a lot of changes in v3. The closest query I have for the purpose is as follows:
```sql
SELECT a.id drug_id, a.name, b.alias cid 
FROM drugs a 
INNER JOIN drug_aliases b ON b.drug_id=a.id
INNER JOIN drug_aliases_sources c ON c.drug_alias_id=b.id
INNER JOIN sources d ON d.id=c.source_id
WHERE d.source_db_name='PubChem' AND b.alias !~ '[^0-9]'
```
Please correct me if anything is wrong. But the where clause is ugly and not strict.

The reason why I dig so hard directly into the postgresql db is that there is no download option available for drug and drug identifier (ChEMBL ID, PubChem CID, or whatever). Moreover, the download option for gene and gene identifier (i.e. Entrez GeneID) seems gone.
[http://dgidb.genome.wustl.edu/downloads](http://dgidb.genome.wustl.edu/downloads)

ahwagner:
Using ChEMBL as our base resource for drugs allows us to specify a ChEMBL ID (although not necessarily a PubChem CID) for each drug. This is available as a field (`chembl_id`) in the `drugs` table.

Similarly, the `genes` table has an `entrez_id` field that designates a unique Entrez identifier for each gene concept.

need47:
Could you add download options for drugs and genes?

ahwagner:
Sure! It sounds to me like you want: 

genes.tsv:
Gene Claim | Gene Group Name | Entrez ID | Gene Claim Source

drugs.tsv:
Drug Claim | Drug Group Name | ChEMBL ID | Drug Claim Source

correct?

need47:
That's good enough for me. Thanks a lot.

need47:
one more suggestion to [interactions.tsv](http://dgidb.genome.wustl.edu/downloads): could you please add drug and gene identifiers?

ahwagner:
That sounds reasonable. We'll work that in.

need47:
Thanks. I never had an experience with such prompt responses like you :-)

ahwagner:
[...]

Issue #158, opened 2017-04-26
Create grouper tsvs


ahwagner:
We currently provide an interactions .tsv. However, in line with expanding our API in #141, we should also endeavor to provide our grouped concepts as .tsv downloads.

PR #237, opened 2017-08-18
Version 2 of the API

### LINKED ISSUE #141: Expose Drug, Gene, Interaction metadata through API
Is there any way to retrieve identifiers through the API for the biological entities in the response? The **interactions** endpoint only returns a `drugName` field, but no identifiers, such as CAS, Drugbank, or Pubchem IDs. 

It looks like these are present in the [database dump](http://dgidb.genome.wustl.edu/downloads) for drugs from at least several of the data sources. I understand data sources often do not provide this information.

Also, I'm wondering if you all have attempted to normalize drugs when a data source does not provide sufficient identifiers. For example, mycancergenome only provides drug names. If a drug is in both mycancergenome and another source, is it treated as two separate entities?

Thanks in advance

susannasiebert:
This includes new endpoints for:
- all interactions/genes/drugs
- one interaction/gene/drug

The interaction search results now also expose more information.

This PR also includes updates to the documentation regarding the changes.

Closes #141, #158, and #218.

### DIFF
--- a/app/models/data_model/interaction_claim.rb
+++ b/app/models/data_model/interaction_claim.rb
@@ -13,5 +13,17 @@ module DataModel
     def self.for_show
       eager_load(:interaction_claim_types, :interaction_claim_attributes, :source, :drug_claim, gene_claim: [:gene, :source])
     end
+
+    def self.for_tsv
+      includes(:source, :interaction_claim_types, drug_claim: [:drug], gene_claim: [:gene])
+    end
+
+    def drug
+      self.drug_claim.drug
+    end
+
+    def gene
+      self.gene_claim.gene
+    end
   end
 end
--- a/app/presenters/interaction_search_results_api_v2_presenter.rb
+++ b/app/presenters/interaction_search_results_api_v2_presenter.rb
@@ -1,4 +1,4 @@
-class InteractionSearchResultsApiPresenter
+class InteractionSearchResultsApiV2Presenter
   def initialize(search_results)
     @search_results = search_results
   end
@@ -6,7 +6,7 @@ class InteractionSearchResultsApiPresenter
   def matched_results
     @matched_results ||= @search_results
       .select { |r| r.is_definite? && r.has_interactions? }
-      .map { |r| InteractionSearchResultApiPresenter.new(r) }
+      .map { |r| InteractionSearchResultApiV2Presenter.new(r) }
   end
 
   def unmatched_results
--- a/app/presenters/interaction_search_results_api_v2_presenter.rb
+++ b/app/presenters/interaction_search_results_api_v2_presenter.rb
@@ -5,13 +5,17 @@ class InteractionSearchResultsApiV2Presenter
 
   def matched_results
     @matched_results ||= @search_results
-      .select { |r| r.is_definite? && r.has_interactions? }
+      .select { |r| r.is_definite? }
       .map { |r| InteractionSearchResultApiV2Presenter.new(r) }
   end
 
   def unmatched_results
     @unmatched_results ||= @search_results
-      .select { |r| r.is_ambiguous? || !r.has_results? }
+      .select { |r| !r.has_results? }
   end
 
+  def ambiguous_results
+    @ambiguous_results ||= @search_results
+      .select { |r| r.is_ambiguous? }
+  end
 end
--- a/app/presenters/interaction_search_results_api_v2_presenter.rb
+++ b/app/presenters/interaction_search_results_api_v2_presenter.rb
@@ -6,7 +6,7 @@ class InteractionSearchResultsApiV2Presenter
   def matched_results
     @matched_results ||= @search_results
       .select { |r| r.is_definite? }
-      .map { |r| InteractionSearchResultApiV2Presenter.new(r) }
+      .map { |r| InteractionSearchResultApiV2Presenter.new(r, r.identifiers[0]) }
   end
 
   def unmatched_results
@@ -17,5 +17,6 @@ class InteractionSearchResultsApiV2Presenter
   def ambiguous_results
     @ambiguous_results ||= @search_results
       .select { |r| r.is_ambiguous? }
+      .flat_map { |r| r.identifiers.map { |i| InteractionSearchResultApiV2Presenter.new(r, i) } }
   end
 end
--- a/app/views/services_v2/interactions.json.jbuilder
+++ b/app/views/services_v2/interactions.json.jbuilder
@@ -1,0 +1,31 @@
+json.matchedTerms @search_results.matched_results do |result|
+  if result.type == 'drugs'
+    json.searchTerm result.search_term
+    json.drugName result.drug_name
+    json.interactions result.interactions do |interaction|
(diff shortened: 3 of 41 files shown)


ahwagner:
Hey @susannasiebert, any idea what's going on with the CI error here?

susannasiebert:
@ahwagner. I pushed a new commit that fixed it. Basically, when I added the search and the entrez_id and chembl_id, I ended up making a second query on the genes/drugs table, respectively. I refactored that functionality to search on the name and id in the same query.

yanyangfeng:
Not sure if this is the place to put this but http://localhost:3000/api/v2/drugs/CHEMBL123702 is returning 404. My db doesn't have this ID and I just pulled latest data dump and did rake dgidb:load_local. The endpoint works for IDs that I do have, so I don't think there's a problem with the API, but we should just update the example route in the API docs to an existing CHEMBL ID.

I know this is not the way it's mean to be used, but http://localhost:3000/api/v2/genes_in_category.json without params returns an error "undefined method `upcase' for nil:NilClass". I just feel like we might want to better handle this exception so people don't get confused?

susannasiebert:
@yanyangfeng I'm doing some digging as to why this ID is in my database and not yours. I also did a `rake dgidb:load_local` and it's still there. Do all of the other examples for the new endpoints work for you?

For the gene_in_categories endpoint I will slightly change the URL in the example, to make clear that it needs to be used with parameters. I don't think we should do anything special with this error though.

susannasiebert:
I actually just typo'ed that chembl id...

susannasiebert:
I pushed a few more commits to update the tsvs.

app/models/data_model/interaction_claim.rb
```diff
@@ -13,5 +13,17 @@ class InteractionClaim < ::ActiveRecord::Base
     def self.for_show
       eager_load(:interaction_claim_types, :interaction_claim_attributes, :source, :drug_claim, gene_claim: [:gene, :source])
     end
+
+    def self.for_tsv
+      includes(:source, :interaction_claim_types, drug_claim: [:drug], gene_claim: [:gene])
```
susannasiebert: @acoffman @ahwagner Is it better here to use `eager_load` instead of `includes` here? I'm not exactly sure in which case to use one over the other.

ahwagner: I took a look at it with `Benchmark.measure`, and it seems to me like `includes` works well for this purpose. Looking at the Rails source code for the two methods ([includes](https://apidock.com/rails/v4.0.2/ActiveRecord/QueryMethods/includes%21), [eager_load](https://apidock.com/rails/v4.0.2/ActiveRecord/QueryMethods/eager_load%21)), it seems to me that eager_load does slightly less compute (no check for blank, no calls to flatten or uniq), but returns cleaner results. 

However, I may be missing some ActiveRecord magic here, as [this post](http://blog.scoutapp.com/articles/2017/01/24/activerecord-includes-vs-joins-vs-preload-vs-eager_load-when-and-where) states that `eager_load` makes fewer queries (1 vs 2 for `includes`), and that is reflected in the call to `for_tsv`. However, since `for_tsv` is only getting called once per export operation, that distinction is moot--both ops performed in ~25-30ms. For this reason, I would stick with `includes`, as it may do a better job of cleaning up the results, and it matches the recommendations in the [Rails Guides](http://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations).

@acoffman, feel free to correct if I'm off base here.

app/presenters/interaction_search_results_api_v2_presenter.rb
```diff
@@ -0,0 +1,17 @@
+class InteractionSearchResultsApiV2Presenter
+  def initialize(search_results)
+    @search_results = search_results
+  end
+
+  def matched_results
+    @matched_results ||= @search_results
+      .select { |r| r.is_definite? && r.has_interactions? }
```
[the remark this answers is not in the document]
susannasiebert: I made these changes as suggested and updated the documentation accordingly.

lib/utils/tsv.rb
```diff
@@ -79,25 +87,65 @@ def self.print_category_row(file_handle, gene)
...
-           gene.name,
-           gene.long_name,
-           interaction.source.source_db_name,
-           interaction.interaction_claim_types.map(&:type).join(','),
-           interaction.drug_claim.name,
-           interaction.drug_claim.primary_name
-        ].join("\t")
+    def self.print_interaction_row(file_handle, interaction_claim)
+      return if license_restricted? interaction_claim.source.source_db_name
+      return unless interaction_claim.interaction_id
```
ahwagner: We will want all interaction claims, not just grouped ones.

app/views/services_v2/interactions.json.jbuilder
```diff
@@ -0,0 +1,37 @@
...
+      json.interactionTypes interaction.types
+      json.drugName interaction.drug_name
+      json.drugChemblId interaction.drug_chembl_id
+      json.sources interaction.source_db_names
+      json.pmids interaction.publications
+    end
+  end
+end
+
+json.ambiguousTerms @search_results.ambiguous_results do |result|
```
ahwagner: `ambiguousTerms` should look like `matchedTerms`, except that `searchTerm` and `drugName`/`geneName` will be different. List full responses for each such suggestion.

lib/utils/tsv.rb
susannasiebert: This has been fixed.

app/views/services_v2/interactions.json.jbuilder
susannasiebert: This has been addressed.

app/views/static/api.html.haml
```diff
@@ -557,7 +1544,7 @@
...
+            curl http://dgidb.genome.wustl.edu/api/v2/interactions.json<br><br>curl http://dgidb.genome.wustl.edu/api/v2/interactions.json?genes=FLT3<br><br>curl http://dgidb.genome.wustl.edu/api/v2/interactions.json?genes=FLT3 | python -mjson.tool<br><br>curl http://dgidb.genome.wustl.edu/api/v2/interactions.json?genes=FLT3,STK1,FAKE1 | python -mjson.tool<br><br>curl http://dgidb.genome.wustl.edu/api/v2/interactions.json?genes=STK1,FAKE1 | python -mjson.tool<br><br>curl http://dgidb.genome.wustl.edu/api/v2/interactions.json?genes=STK1,FAKE1 | python -mjson.tool<br><br>curl http://dgidb.genome.wustl.edu/api/v2/interactions.json?genes=FLT3,STK1,FAKE1\&interaction_sources=TTD,DrugBank | python -mjson.tool<br><br>curl http://dgidb.genome.wustl.edu/api/v2/interactions.json?genes=FLT3,STK1,FAKE1\&interaction_sources=TTD,DrugBank\&interaction_type=inhibitor | python -mjson.tool<br>
```
ahwagner: The `\&`s are interpreted literally here, resulting in a `\` displaying on page. Consider using `&amp;` instead.

PR #241, opened 2017-08-25
New online updater for CKB

susannasiebert:
This PR also adds a parent class to be used by the online updaters that includes methods for creating the different database entries for drug/gene/interactions claims and their aliases and attributes.


ahwagner:
Looks like the test suite can't find `Genome::OnlineUpdater`. @susannasiebert, any idea what's causing this?

susannasiebert:
The tests are passing for me locally, so I'm not sure... The travis tests were both failing with `< Genome::OnlineUpdater` and `< OnlineUpdater`. Both versions pass for me locally. Any idea how to make travis "see" this module/class?

susannasiebert:
The `require` statements did the trick. Tests are passing now.

Issue #239, opened 2017-08-22
remove non-curated vs expert curated coloring


yanyangfeng:
Make the source tags all one color in interaction search results

Issue #238, opened 2017-08-22
Link directly from interaction search results to interaction views


yanyangfeng:
Right now, the tables in the interaction search results panels can only link directly to the genes/drugs that interact with your matched search terms. If you want to view the full interaction view, you have to click on the gene/drug, then go its interactions tab, then find the term you originally searched for. Linking directly from interaction search results, say by clicking on the table row, would eliminate this circuitous process.

Issue #152, opened 2017-04-21
add WUSTL and MGI branding to site


ahwagner:
from #151

yanyangfeng:
@malachig, @ahwagner, was there anything I needed to do to make sure this branding stuff wouldn't piss any admin off?

malachig:
Can you post a screenshot for reference?

yanyangfeng:
![screen shot 2017-08-25 at 1 40 49 pm](https://user-images.githubusercontent.com/16638655/29727855-37e16708-899b-11e7-8152-ce8992e37d01.png)

malachig:
This looks fine to me.

PR #242, opened 2017-08-25
Branding, link interaction results to interactions directly, same color for expert and non-curated

yanyangfeng:
Addresses #152, #238, #239


ahwagner:
It looks like there are some failing tests with respect to the gene categories page? @yanyangfeng, would you look into this?

yanyangfeng:
Yeah, accidentally left a trustlevel presenter method commented out, checks passing now.

PR #245, opened 2017-08-29
Page headers

yanyangfeng:
removed page headers and added label/icon to drug/gene/interaction view summary panels


ahwagner:
Hey @yanyangfeng, would you look into the failing tests here? Also, you'll need to merge in the current `master` branch and resolve conflicts in the files listed above.

yanyangfeng:
So it seems that these checks are failing because the tests check for certain text on each page to prove that the page has loaded, and this text happens to be in the headers. I could modify the tests to check for the presence of something else uniquely identifying to each page, or is there some other way of going about this?

ahwagner:
Looking for different, unique content makes sense to me.

yanyangfeng:
I reduced the space a bit - the search result tabs do butt into the background image a bit, but I think it's actually kind of a cool effect.

PR #257, opened 2017-09-05
CGI online updater

kcotto:


ahwagner:
hey @kcotto, how is progress on this PR?

kcotto:
So when I check the data.sql file after running the importer, I see this in the sources table. So it looks like the source is being created but the claims aren't/ are not being saved. 

> CGI	11-September-2017	https://www.cancergenomeinterpreter.org/	http://https://www.cancergenomeinterpreter.org/biomarkers	https://www.cancergenomeinterpreter.org/	Cancer Genome Interpreter	ce28974a-b813-4717-a3d2-2df0cc54ab5a	0	0	0	0	0	0	\N	0

ahwagner:
@kcotto did you run `pg = PostGrouper.new; pg.perform` after grouping? That will update the source table to the current counts.

susannasiebert:
The post grouper should've been run as part of the `UpdateCgi`'s `perform` call (which is inherited from the `TsvUpdater`), unless you ran the `NewCgi` importer by itself.

kcotto:
I reran the importer and it looked like the grouper ran. Still no claims though.

ahwagner:
@kcotto is this ready for review?

kcotto:
Working on one more piece and then it will be done

kcotto:
A few lines are have weird characters/phrases that are interfering with drug grouping but otherwise this is good.

Issue #255, opened 2017-09-02
add empty box to results page if no results


ahwagner:
currently, if there are no unique matches or ambiguous matches, the respective tabs of the interaction search results page have no results, but the sidebar boxes remain, creating a weird effect: 

![screen shot 2017-09-01 at 9 30 44 pm](https://user-images.githubusercontent.com/811065/29992257-df562ffc-8f5c-11e7-8957-a912be9e969c.png)

We should fix that by adding in an empty result box declaring "No unique (ambiguous) results" in these situations.

Issue #244, opened 2017-08-29
Custom source cards for Entrez and ChEMBL


ahwagner:
Entrez has a source in our database right now, but no gene claims (because they are the base concept of `gene` and are treated differently). The existing ChEMBL source should be renamed to ChEMBL interactions, and a new ChEMBL source should be created for drugs. These two sources (ChEMBL drugs and Entrez genes) will need custom methods written for them in the `PopulateCounters` normalizer.

ahwagner:
Remove genes/drugs panels from sources as appropriate based on source type. Remove trust information panel. Separate out interactions into a new sub-panel.

susannasiebert:
Also discussed: remove inter-gene interactions counter

yanyangfeng:
@ahwagner is someone working on the actual creation of the new sources, or should I tackle that in addition to the normalizer task/views?

ahwagner:
The Entrez source exists already, but the counters are all zeros because as the canonical source there are no associated gene claims. You can start by adjusting the count normalizer to populate the "gene claims" and "grouped claims" to be equal to the total number of genes (DataModel::Gene.count).

The ChEMBL data exist, but no corresponding DataModel::Source exists in the database. I will update and push that change shortly.

ahwagner:
@yanyangfeng this is now merged into `master`, commit c6aafe63bc4dfff24e9ba9c441a2461f710dcf3a. You'll need to update the 'ChemblDrugs' source in the normlizer to reflect the total number of drug groups.

PR #258, opened 2017-09-05
Icons

yanyangfeng:
addresses #255 and #244


yanyangfeng:
how do I deal with the data merge conflict?

ahwagner:
That depends on what is in your data commit, but assuming it's just the additional chembl source and updated counts on sources, I suggest that you merge in `master`, and keep the data commit from `master` (`84bf5ba`). From there, I would run through the normalizer, ensure that the counts are populating correctly, and create a new data commit on top of `84bf5ba` with the populated count fields.

ahwagner:
Nevermind, the data commit is an easy fix--I'm pushing those changes presently.

PR #259, opened 2017-09-05
Tree menu

yanyangfeng:
pretty old changes that I think I just never pushed - changes the multiple matches sidebar so that there's a clear visual mapping of each mapped entity to its search term


ahwagner:
This looks pretty good. Let's cut down the margin left of the term list (red box, below) to leave a little more space for gene symbols.

screen shot 2017-09-05 at 2 11 46 pm

yanyangfeng:
Left aligned with Multiple Matches heading text

PR #264, opened 2017-09-11
New online updater for Guide to Pharmacology

susannasiebert:
This online updater is currently very slow since it has to make multiple API calls for each claim we want to create. I'm happy to incorporate any suggestions as to how to speed it up, if possible.

I've tried running it twice so far and both times it failed about 5-6 hours in because a API request failed. From the description it isn't clear to me if it failed because of some intermittent server issue or what.

Alternatively, we could go back to using the tsv importer and wrapping it in our online updater functionality (probably rewriting the tsv importer in the process).

### DIFF
--- a/app/jobs/update_guide_to_pharmacology.rb
+++ b/app/jobs/update_guide_to_pharmacology.rb
@@ -1,0 +1,32 @@
+class UpdateGuideToPharmacology < TsvUpdater
+  def tempfile_name
+    ['interactions', '.tsv']
+  end
+
+  def importer
+    Genome::Importers::GuideToPharmacologyInteractions::NewGuideToPharmacology.new(tempfile)
+  end
+
+  def next_update_time
+    Date.today
+      .beginning_of_week
+      .next_month
+      .midnight
+  end
+
+  def latest_url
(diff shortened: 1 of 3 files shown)


susannasiebert:
Part of the reason that this is so long-running is that there are over 13000 interactions to import.

ahwagner:
Hrm. I don't like the idea of hammering the server with multiple requests for each of 13K interactions. I agree that we should go to an online tsv importer.

susannasiebert:
I changed this updater to be tsv-based. It imports fine but drug grouping fails. @ahwagner would you have time to look into that? It didn't see anything obvious.

susannasiebert:
I think I just needed to run `rake dgidb:load_local`.

susannasiebert:
This are the stats after grouping:
```
 gene_claims_count: 1608,
 drug_claims_count: 6495,
 interaction_claims_count: 12593,
 interaction_claims_in_groups_count: 7876,
 gene_claims_in_groups_count: 1573,
 drug_claims_in_groups_count: 3822,
```

ahwagner:
👍

PR #266, opened 2017-09-12
Size fixes

yanyangfeng:
- "anti-neoplastic" to "antineoplastic"  in icons and filter text 
- header now scrolls to make tabs accessible in small window
- remove loading modal in safari 
- add margins to source icons in interaction search panel tables to fix overlap in small window
- adjust anchor compensation in api page

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 gem 'rails', '5.1.1'
 
+gem 'browser'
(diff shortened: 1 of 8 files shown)


yanyangfeng:
just pushed changes to fix the typo on the API page (v1 and current) and added a max-width to css of application layout to fix hidden footer

ahwagner:
👍

PR #272, opened 2017-09-15
Size fixes

yanyangfeng:
- dgidb.org instead of dgidb.genome.wustl.edu on api docs (v1 and current)
- fix n/a interaction types in summary views
- center search forms (interactions and categories)


yanyangfeng:
just pushed: created stats page which for now is just a well that lists total interaction, total gene, total drug, and total druggable gene counts and added a link to the info drop down menu.

PR #275, opened 2017-09-15
Update DrugBank XML parsing of references

### LINKED ISSUE #274: DrugBank claims have PMID 0
Strange, but somehow all 11293 drugbank interaction claims have an associated PMID of 0. I've cleaned up the database, but we should check the importer to see where things are going wrong.

susannasiebert:
Closes #274.


susannasiebert:
@ahwagner This should now be fixed.

Issue #273, opened 2017-09-15
high failure rate of drug grouping from sources


ahwagner:
Our global ungrouped rate for drugs is ~25%. Some sources are bigger offenders than others:

- [ ] CKB (865, 73%) (!!)
- [ ] GuideToPharmacologyInteractions (2450, 38%)
- [ ] DrugBank (1636, 33%)
- [ ] NCI (517, 34%)
- [ ] PharmGKB (194, 33%)
- [ ] ClearityFoundationClinicalTrial (30, 26%)
- [ ] CIViC (73, 26%)

For reasons discussed in the paper, there will be a number of drug claims that simply won't group to a single drug entity--that's unavoidable. However, I feel like there is still room for improvement here, which we should strive to accomplish before the revised paper submission.

susannasiebert:
I'm investigating CKB and it seems to be because we are importing drugs without consideration for combination therapies. I will update the CKB api updater accordingly.

ahwagner:
investigate extending NCIT drug normalization from CIViC

ahwagner:
I've pushed my work on this up to https://github.com/cancervariants/therapy-normalization/blob/master/analysis/DGIdb.ipynb.

I've started with a Wikidata normalizer. Long story short: there's a long way to go.

susannasiebert:
The new grouping percentages using the updated drug grouper are:
```
TTD: 21.54%
GuideToPharmacology: 31.79%
DrugBank: 44.98%
NCI: 46.81%
JAX-CKB: 54.43%
CIViC: 58.86%
ClearityFoundationClinicalTrial: 63.48%
TdgClinicalTrial: 71.72%
TALC: 72.63%
MyCancerGenomeClinicalTrial: 73.21%
CGI: 73.55%
PharmGKB: 76.59%
ClearityFoundationBiomarkers: 79.69%
CancerCommons: 87.5%
FDA: 87.97%
COSMIC: 89.29%
OncoKB: 89.89%
MyCancerGenome: 90.08%
TEND: 92.01%
DoCM: 94.87%
ChemblInteractions: 98.71%
DTC: 99.9%
ChemblDrugs: 100.0%
Wikidata: 100.0%
```

susannasiebert:
Overall our grouping percentage is 45.91%. I think the biggest problem is TTD as we are importing way more drug claims (28,644) from this source now than we used to.

susannasiebert:
[...]

PR #277, opened 2017-09-15
Fix CKB updater to account for combination therapies

susannasiebert:
My local run hasn't finished grouping yet but this should improve the drug grouping success rate for CKB (see #273). I will update with stats once the grouper finishes.

### DIFF
--- a/lib/genome/online_updaters/ckb/updater.rb
+++ b/lib/genome/online_updaters/ckb/updater.rb
@@ -38,10 +38,22 @@ module Genome; module OnlineUpdaters; module Ckb;
         gene_claim = create_gene_claim(gene['geneName'], 'CKB Gene Name')
         create_gene_claim_aliases(gene_claim, gene)
         api_client.interactions_for_gene_id(gene['id']).each do |interaction|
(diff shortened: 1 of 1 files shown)


susannasiebert:
I finished grouping and made the new database dump part of this PR. The new stats for this source are:
```
 gene_claims_count: 86,
 drug_claims_count: 729,
 interaction_claims_count: 1949,
 interaction_claims_in_groups_count: 1411,
 gene_claims_in_groups_count: 86,
 drug_claims_in_groups_count: 420,
```
So the drug claim grouping success rate is slightly better at 57.6%. Still not all that great so further investigation will be necessary.

susannasiebert:
Part of the reason for the low grouping % is that we don't import any aliases from CKB. In order to do so we would need to scrape an additional web page, one for each drug, e.g., https://ckb.jax.org/therapy/show/5463. We need to decide wether the additional import overhead is worth it to achieve a higher grouping percentage.

ahwagner:
I'll see if we can get somebody from the CKB to give us that info as part of a single endpoint.

PR #297, opened 2017-09-29
Strip names/attributes/aliases/etc of leading and trailing whitespaces

### LINKED ISSUE #293: DoCM online updater imports drug claim names with leading spaces.
I'm not sure if this needs to be fixed in DoCM or handled by us but the DoCM online updater imports some claim names with a leading whitespace.

Example:
```
#<DataModel::DrugClaim:0x007fe6fbb67710
 id: "070150d6-5fa2-4328-ae48-fc050254052b",
 name: " PKC412",
 nomenclature: "DoCM Drug Name",
 source_id: "b3521d9b-ba98-48b3-9834-e63191146367",
 primary_name: " PKC412",
 drug_id: "ee7f7a80-e97d-4968-9842-07d559733b6e">
```
We should update the online updater to clean up these values before importing.

susannasiebert:
Closes #293.

This also fixes this issue with the PubMed scraping. This PR changes it so that we are using the official API.

### DIFF
--- a/lib/utils/database.rb
+++ b/lib/utils/database.rb
@@ -258,6 +258,14 @@ module Utils
         a.value = a.value.strip
         a.save!
       end
+      DataModel::InteractionAttribute.where("name LIKE ' %' or name LIKE '% '").all.each do |a|
+        a.name = a.name.strip
+        a.save!
+      end
+      DataModel::InteractionAttribute.where("value LIKE ' %' or value LIKE '% '").all.each do |a|
+        a.value = a.value.strip
+        a.save!
+      end
 
       DataModel::GeneClaimAlias.where("alias LIKE ' %' or alias LIKE '% '").all.each do |a|
         a.alias = a.alias.strip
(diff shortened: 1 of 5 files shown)


lib/utils/database.rb
```diff
@@ -244,5 +244,98 @@ def self.model_to_tsv model
         end
       end
     end
+
+    def self.cleanup_whitespaces
+      DataModel::InteractionClaimType.where("type lIKE ' %' or type LIKE '% '").all.each do |t|
```
yanyangfeng: does the L need to be capitalized?

lib/utils/pmid.rb
```diff
@@ -20,89 +20,65 @@ def self.pubmed_url(pubmed_id)
...
-          last_name: author.xpath('LastName').text,
-          author_position: i
-        }
-      end
+      result['authors'].map{|a| a['name']}
     end
 
     def pmc_id
-      xpath_contents_or_nil("//PubmedData/ArticleIdList/ArticleId[@IdType='pmc']")
+      result['articleids'].each do |articles|
```
yanyangfeng: [16] pry(main)> response.pmc_id
=> [{"idtype"=>"pubmed", "idtypen"=>1, "value"=>"1"},
 {"idtype"=>"rid", "idtypen"=>8, "value"=>"1"},
 {"idtype"=>"eid", "idtypen"=>8, "value"=>"1"}]

I feel like we need a return nil at the end of this function, in the case where there is no pmc_id

lib/utils/pmid.rb
```diff
@@ -20,89 +20,65 @@ def self.pubmed_url(pubmed_id)
...
     def year
-      xpath_contents_or_nil('//Journal/JournalIssue/PubDate/Year')
-    end
-
-    def month
-      monthname = xpath_contents_or_nil('//Journal/JournalIssue/PubDate/Month')
-      if monthname
-        Date::ABBR_MONTHNAMES.index(monthname)
+      if Date.new(publication_date.to_i).gregorian?
+        publication_date
```
yanyangfeng: this should just be publication_date.to_i if we want to extract only the year, but won't all our years be gregorian?

https://ruby-doc.org/stdlib-2.4.1/libdoc/date/rdoc/Date.html#method-i-gregorian

susannasiebert: This doesn't quite seem to do what I wanted it to. Sometimes the date in this field is just the year and this if-clause was supposed to catch that. I will rework this.

lib/utils/database.rb
susannasiebert: I don't think capitalization really matters but I pushed a commit to fix it.

susannasiebert:
@yanyangfeng I pushed a few commits to address your comments.

susannasiebert:
I think I might need to do a regrouping on the cleaned data. I see that the claims are fixed but the groups still have this problems. I thought I already regrouped but apparently not.

susannasiebert:
I just committed a new data dump that has the trailing/leading whitespaces of group attributes and aliases fixed. That data dump also contains cleaned up PharmKGB drug (claim) attributes.

ahwagner:
@susannasiebert did you do a full regroup after?

susannasiebert:
I did not, mainly because I don't think we delete existing group attributes and aliases during regrouping. That's why I think it was necessary to fix the group level attributes/aliases directly.

ahwagner:
Ah, this makes sense. Misread https://github.com/griffithlab/dgi-db/pull/297#issuecomment-334776492, now I understand.

PR #299, opened 2017-10-03
Category view

yanyangfeng:
addresses #283

### DIFF
--- a/app/controllers/genes_controller.rb
+++ b/app/controllers/genes_controller.rb
@@ -34,6 +34,7 @@ class GenesController < ApplicationController
   end
 
   def categories_search_results
+    @view_context = view_context
     @search_categories_active = 'active'
     start_time = Time.now
     validate_search_request(params)
--- a/app/presenters/gene_category_search_results_presenter.rb
+++ b/app/presenters/gene_category_search_results_presenter.rb
@@ -80,7 +80,7 @@ class GeneCategorySearchResultsPresenter
   def unique_matches
     unless @unique_matches
       @unique_matches = []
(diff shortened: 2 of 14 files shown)


kcotto:
Something weird is happening on the by category page. It appears it's not showing all of the druggable gene categories that have a result in them. The live site shows 16 druggable gene categories but I'm only seeing 11 on this when viewing that tab on this PR.

Also, do we want to consider sorting the druggable gene categories either alphabetically or ordering by the number of genes from the query within them?

yanyangfeng:
I'll sort the by category results by number of genes and leave browse categories in alphabetical order.

I'm currently looking into the by category issue - I'm seeing the same thing, but I'm wondering if this is a database thing?

yanyangfeng:
seems like there is a view bug and just an underlying difference in the database.

yanyangfeng:
The visual bug is fixed, pushing commit to this PR. 

There doesn't appear to be anything funky going on, assuming my db is in the state we want it. I went through psql and ran SELECT * FROM genes INNER JOIN gene_categories_genes ON genes.id = gene_categories_genes.gene_id INNER JOIN gene_claim_categories on gene_categories_genes.gene_claim_category_id = gene_claim_categories.id WHERE genes.name=<gene> for each of the uniquely matching genes in the example search and it looks like all the categories are being represented.

kcotto:
screen shot 2017-10-06 at 3 39 05 pm

I can see all of the results now. However, I'm still not seeing any sort of sorting on this tab, either by alphabetically or by the number of results.

PR #307, opened 2017-10-17
Fix API bug and documentation and add more tests

### LINKED ISSUE #306: drug gene interactions endpoint
API lookups for specific interactions fail, e.g.:
http://dgidb.org/api/v2/interactions/0031edb6-06a9-4938-98c7-147ba88d190f

We should write a test for this in the patch.

susannasiebert:
Closes #306


ahwagner:
Looks good on local. Testing in staging env now.

PR #308, opened 2017-10-18
Home page and info menu/tabs to dropdown menus

yanyangfeng:
Addresses comments 1-3 by the reviewer


susannasiebert:
Since Downloads doesn't have any other sub-topics, I suggest that we make that tab a link directly to the download page instead of a dropdown menu to only one option.

yanyangfeng:
@susannasiebert don't have any huge qualms with that - only thing I was thinking was that if people were browsing through the tabs, thinking they were all dropdowns, they would be unexpectedly thrown into a new page. Obviously not a huge deal but I thought it might be better to keep things consistent. Thoughts?

susannasiebert:
Good point. What if we made two options: `data` and `source code` with the former going to our downloads page and the latter going directly to GitHub?

yanyangfeng:
Yeah I think that's a good idea. If no one else objects, I think I'll go with that, and remove the repo link that currently exists in the downloads page

yanyangfeng:
@ahwagner so I removed the 80% zoom because I thought the reviewer had complained about how the text seemed very small everywhere on the page. That, and there's a weird issue with the zoom in firefox where it literally zooms out the browser screen and there's a ridiculous white border around the page. Something we want to push back on?

ahwagner:
I think that the interaction search results are the only place where font might be too small on the current live site. After disabling 80% zoom, these are fine, but everything else seems way too big. Perhaps we can reduce the `font-size` of `h1`, `h2`, `h3` elements?

ahwagner:
Ugh. Reducing those elements didn't help much--images / banner are still much too large. Let's leave the 80% zoom, and figure out the ugly zoom on Firefox bug later.

yanyangfeng:
I think my most recent push addresses all of these concerns. Let me know if I missed anything. For the zoom, I left out the firefox-specific scale css that was making it ugly, so now it's merely bloated in firefox. I added zoom 80% in general, and specifically for the interaction search results, I added zoom 125% to cancel this out.

yanyangfeng:
oops, should probably zoom the categories search results as well

PR #316, opened 2017-11-22
Fix /html error in travis 

kcotto:


kcotto:
Indentation error for style guidelines. For future note, everything must be indented at least one more than the html wrapper. The code outside of the </html> was what was failing the Travis test.

PR #319, opened 2018-01-23
Analytics fix

kcotto:


ahwagner:
Looking back through our commit history, it looks like there has been quite a bit of modification to `application.html.haml` that was erroneous duplication and tabbing, which has created this issue. Historically, we have had this code at the bottom of the document (not in the header), and it had worked properly. I have cleaned up these code issues on a different PR (https://github.com/griffithlab/dgi-db/pull/320), reverting the structure to a state similar to how it worked previously and am closing this PR accordingly.

PR #346, opened 2019-08-21
new importer for tempus

sharonfreshour:
Need to add tsv file
Run importer locally, verified it imported correctly


ahwagner:
Cool! @sharonfreshour don't forget to move this from "draft" to "ready for review" when you think it's ready!

susannasiebert:
@sharonfreshour just pushed an update to the data submodule to include the tempus tsv. Basically, we cd'ed into the data folder, added and committed the tsv, cd'ed to the main folder and added and committed the data folder/submodule. It's been a while since I have done this so if someone can confirm that these are all the necessary steps, that would be great!

kcotto:
I might be remembering this incorrectly but don't you also have to commit within the data folder since it is its own git repo?

susannasiebert:
@kcotto, yes we did that.

sharonfreshour:
No, I don't think that was supposed to be included. I'll fix that.

sharonfreshour:
I think the commit should be pushed to git.genomedata.org now

lib/genome/importers/tempus/tempus_tsv_importer.rb
```diff
@@ -0,0 +1,27 @@
+module Genome
+  module Importers
+    module Tempus
+      def self.source_info
+        {
+          base_url:           'https://www.tempus.com/',
+          site_url:           'https://www.tempus.com/',
+          citation:           'https://www.tempus.com/',
```
ahwagner: It would be nice (if available) to link to a publication of theirs. Are there any from https://www.tempus.com/publications/ that describe the gene list? If not, perhaps we can cite the publications page here instead of the homepage.

lib/genome/importers/tempus/tempus_tsv_importer.rb
```diff
@@ -0,0 +1,27 @@
+module Genome
+  module Importers
+    module Tempus
+      def self.source_info
+        {
+          base_url:           'https://www.tempus.com/',
```
ahwagner: This will require updates to the following codeblocks: 
https://github.com/griffithlab/dgi-db/blob/d41012579be0e7d3db9ff96b2b785201491faa64/app/models/data_model/gene_claim.rb#L66-L67 and https://github.com/griffithlab/dgi-db/blob/d41012579be0e7d3db9ff96b2b785201491faa64/app/models/data_model/drug_claim.rb#L52-L53

lib/genome/importers/tempus/tempus_tsv_importer.rb
sharonfreshour: There is a publication about validation of this panel: https://www.tempus.com/clinical-validation-of-the-tempus-xt-next-generation-targeted-oncology-sequencing-assay/

This would be the link to use for the citation link, correct? Should anything be changed in the base or site urls (aside from the codeblocks mentioned below)?

sharonfreshour:
I think all the necessary changes have been made

Issue #348, opened 2019-08-22
Clean up unused files and dirs


acoffman:
Some files and directories date back 7+ years and aren't used any longer. Delete them.

PR #349, opened 2019-08-22
Start removal of unused files and directories

acoffman:
This removes

* `debian/` which contained the old packaging scripts for making `.debs` before we switched to Capistrano and AWS
* `doc/` which was an autogenerated directory that was never used
* `dgidb-apache.conf` which was an apache config file that we don't use that dates back to the old debian packaging
* `gmt.yml` and `gmt/` which were config files for the `gmt.genome.wustl.edu` site generator which is no longer maintained.
* `vendor/` which was an autogenerated directory from an old version of rails and not in use.


ahwagner:
Supports issue #348

PR #360, opened 2019-11-04
Calculate a more comprehensive score for interaction search results

### LINKED ISSUE #340: Query overlap score
If multiple queried genes match to a drug, we should indicate that in some way

susannasiebert:
Closes #340

### DIFF
--- a/app/presenters/interaction_search_results_presenter.rb
+++ b/app/presenters/interaction_search_results_presenter.rb
@@ -36,30 +36,21 @@ class InteractionSearchResultsPresenter
   end
 
   def scores_for_result(result, grouped_results)
-    if @search_context == "genes"
-      scores = result.interactions.values.flatten.each_with_object({}) do |result_interaction, h|
-        overlap_count = grouped_results.map{ |r|
-          r.interactions.values.flatten.count{|other_interaction| other_interaction.drug_id == result_interaction.drug_id}
-        }.sum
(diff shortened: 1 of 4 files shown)


app/presenters/interaction_search_results_presenter.rb
```diff
@@ -35,6 +35,34 @@ def definite_no_interactions
     results_for_view(Maybe(grouped_results[:definite_no_interactions]))
   end
 
+  def scores_for_result(result, grouped_results)
+    if @search_context == "genes"
+      scores = result.interactions.values.flatten.each_with_object({}) do |result_interaction, h|
+        overlap_count = grouped_results.map{ |r|
+          r.interactions.values.flatten.count{|other_interaction| other_interaction.drug_id == result_interaction.drug_id}
+        }.sum
+        promiscuity_count = DataModel::Interaction.where(drug_id: result_interaction.drug_id).count
```
acoffman: Are the only things differing between the two outer blocks the `drug_id` here and on 47 vs the `:gene_id` below?

If so, would it make more sense to do something like

```
identifiier  = if @search_context == 'genes'
   :drug_id
else
  :gene_id
```

and then just have the rest of the calculation be outside the conditional with `DataModel::Interaction.where(identifier => result_interation.send(identifier))`?

susannasiebert:
@acoffman I removed the code duplication as suggested. If you could have another look and re-approve, that would be great.

PR #366, opened 2020-01-09
IDG importer

sharonfreshour:
Pull request for new idg tsv importer


app/presenters/interaction_presenter.rb
```diff
@@ -11,10 +11,14 @@ def display_name
   end
 
   def na_or_types
-    types = interaction.interaction_types.map{|t| t.type}.reject{|t| t == "n/a"}
+    types = interaction.interaction_types.map{|t| t.type.downcase}.uniq.reject{|t| t == "n/a"}
```
susannasiebert: This will result in this method returning the downcased version of the type. I assume that is to ensure that the uniq really catches all cases, irregardless of capitalization. However, I would argue that we should instead ensure that the values entered in to the database aren't just differently capitalized duplicates.

sharonfreshour: @susannasiebert Would that involve making sure that importers downcase interactions when they are imported? Or would that be addressed some where else?

susannasiebert: I think the first thing to decide is if downcasing is what we want or something fancier (e.g. `capitalize` with a special case for `N/A`). I would then suggest to create a function for that conversion in the model for the interaction type which can then be called by all the importers when an interaction type gets created by the importer.

lib/genome/importers/idg/idg_tsv_importer.rb
```diff
@@ -0,0 +1,27 @@
+module Genome
+  module Importers
+    module IDG
+      def self.source_info
+        {
+          base_url:           'https://druggablegenome.net/index/',
+          site_url:           'https://druggablegenome.net/index/',
+          citation:           'https://druggablegenome.net/IDGProteinList/',
```
kcotto: I would switch the base url and the citation url for to be more consistent with other importers.

lib/genome/importers/tempus/tempus_tsv_importer.rb
```diff
@@ -0,0 +1,27 @@
+module Genome
+  module Importers
+    module Tempus
+      def self.source_info
+        {
+          base_url:           'https://www.tempus.com/',
+          site_url:           'https://www.tempus.com/',
+          citation:           'https://www.tempus.com/clinical-validation-of-the-tempus-xt-next-generation-targeted-oncology-sequencing-assay/',
```
kcotto: I would switch the base url and the citation url for to be more consistent with other importers.

sharonfreshour:
I think I covered all of the changes discussed, however I am having an issue with the IDG importer still, so I was hoping @kcotto and/or @susannasiebert could take a look at it. I keep getting an issue when I run the importer in rails saying the rows don't match the specified attributes.

PR #376, opened 2020-04-16
closes #375

### LINKED ISSUE #375: Upgrade Ruby version
Currently, our CI and install directions use Ruby 2.3, which is [beyond EOL](https://github.com/rbenv/ruby-build/issues/1353). This causes install errors on OS X, as OpenSSL 1.0 support has been dropped in homebrew as a result of this EOL, and OpenSSL 1.1 is incompatible with Ruby <2.4.

We need to upgrade Ruby, ideally to the most recent version compatible with Rails 5.

ahwagner:
Updates are deployed on staging: http://34.209.88.170

Feels good to have all of this updated. No major changes, just versioning and associated config, should only require one 👍 .


ahwagner:
Re-enabled Travis, all tests passing!

config/deploy.rb
```diff
@@ -13,3 +13,4 @@
 
 set :linked_files, fetch(:linked_files, []).push('config/database.yml', 'config/secrets.yml')
 set :linked_dirs, fetch(:linked_dirs, []).push('log', 'tmp/pids', 'tmp/cache', 'tmp/sockets', 'vendor/bundle', 'public/system', 'public/uploads', 'public/data')
+append :linked_dirs, '.bundle'
```
acoffman, line 16: I am not 100% sure what this line does, and we don't seem to be using  it in other projects. Is it needed?

ahwagner: This is a capistrano optimization. Some background on when it was introduced and why here: https://github.com/capistrano/bundler/issues/96

PR #377, opened 2020-04-27
Normalize interaction claim types

### LINKED ISSUE #367: Normalize Interaction Types
Right now we have duplicate InteractionClaimTypes in the database that only differ in their capitalization. We should normalize them:

- Make a list of the existing values to ensure that there aren't any values we don't want
- Lowercase the existing values, remove duplicate values and reassign linked interactions/interaction claims to the remaining value
- Add some database constraints to the `type` column to a) enforce lowercasing and b) enforce unique values
- Update the importers to not create new InteractionClaimTypes and instead fail if an interaction type is not found

The `downcase` and `uniq` calls in https://github.com/griffithlab/dgi-db/pull/366 should be removed.

This might be related to issue: https://github.com/griffithlab/dgi-db/issues/202

susannasiebert:
Closes #367


lib/genome/normalizers/interaction_claim_type.rb
```diff
@@ -42,12 +61,34 @@ def self.add_unless_exists(type, interaction_claim)
...
+
       def self.name_normalizer(val)
         val = val.downcase
         if val == 'na' || val == 'n/a'
           'n/a'
         elsif val =~ /other/ || val =~ /unknown/
           'other/unknown'
+        elsif val == 'neutralizer' || val == 'reducer' || val == 'metabolizer' || val == 'acetylation' || val == 'chelator' || val == 'cross-linking/alkylation' || val == 'positive modulator'
+          'modulator'
+        elsif 'inhibitor, competitive' || val == 'gating inhibitor'
```
ahwagner, line 78:
```suggestion
        elsif val == 'inhibitor, competitive' || val == 'gating inhibitor'
```

lib/genome/normalizers/interaction_claim_type.rb
```diff
@@ -42,12 +61,34 @@ def self.add_unless_exists(type, interaction_claim)
...
+        end
+      end
+
       def self.name_normalizer(val)
         val = val.downcase
         if val == 'na' || val == 'n/a'
           'n/a'
         elsif val =~ /other/ || val =~ /unknown/
           'other/unknown'
+        elsif val == 'neutralizer' || val == 'reducer' || val == 'metabolizer' || val == 'acetylation' || val == 'chelator' || val == 'cross-linking/alkylation' || val == 'positive modulator'
```
ahwagner, line 76: @kcotto we are normalizing many concepts to `negative modulator`, but here we're normalizing `positive modulator` to `modulator`. Why are we stepping back from using `positive modulator` as a normalized concept?

susannasiebert:
I converted this to a draft PR so that we don't merge it before I can update the data dump.

lib/genome/normalizers/interaction_claim_type.rb
kcotto: The only concept of a postitive modulator we currently have in our table is `positive allosteric modulator`. I just wasn't sure if we wanted to add `positive modulator` to the table as well. If we do, I can totally make that change.

ahwagner: I think that we should keep to the notion of a `positive modulator` where we can, though it is appropriate to normalize to an unspecified modulator type if we think it is neutral, or do not have enough information to normalize to negative / positive. In general, I think that keeping to the positive/neutral/negative definitions [as described on Wikipedia](https://en.wikipedia.org/wiki/Allosteric_modulator) is safe.

susannasiebert: So `positive modulator` should be kept as such. Should `positive allosteric modulator` then also be normalized to `positive modulator`? I don't see it in the list currently.

kcotto: Yeah, I think the plan was to normalize `positive allosteric modulator` to `positive modulator` to keep consistent with also normalizing concepts to `negative modulator`

susannasiebert: This change has been implemented.

PR #386, opened 2020-05-14
closes #385

### LINKED ISSUE #385: Add directionality to data model
It would be useful to be able to get the directionality of an interaction directly from the interaction model, instead of the corresponding presenter.

ahwagner:
Interaction directionality should be included at the model level, not the presenter.


app/models/data_model/interaction.rb
```diff
@@ -23,5 +23,9 @@ def type_names
     def pmids
       self.publications.pluck(:pmid).uniq
     end
+
+    def directionality
+      self.interaction_types.pluck(:directionality).reject{ |d| d.nil? }.uniq
```
acoffman, line 28: This will work fine as written and presumably one interaction won't have tons and tons of types so the performance impact is probably negligible, but if you wanted you could rewrite this is 

`interaction_types.where.not(directionality: nil).distinct.pluck(:directionality)` and that would do everything in the database instead of pulling everything back and then iterating over it 2x to do the `reject` and the`uniq`. Alternately if you want to stick with this approach, ruby collections have `array.compact` which returns a copy with the `nils` removed for you.

ahwagner: Thanks @acoffman! Implementing the in-database solution you recommended.

PR #388, opened 2020-05-14
Create NDEX tsv

ahwagner:
This PR introduces a new, simplified .tsv for use with NDEX. It adds eager loading to the Interaction class for tsv generation and clarifies the use of interaction claims to generate the current interactions.tsv.


ahwagner:
Yes, we should generate this with the same frequency that we generate the other TSVs.

PR #390, opened 2020-05-27
Set up sidekiq for background job management

susannasiebert:


app/jobs/api_updater.rb
```diff
@@ -1,16 +1,12 @@
...
-      pre_grouper = PreGrouper.new
-      pre_grouper.perform
-      grouper = Grouper.new
-      grouper.perform(should_group_genes?, should_group_drugs?)
-      post_grouper = PostGrouper.new
-      post_grouper.perform(should_cleanup_gene_claims?)
-    ensure
-      reschedule if recurring
-    end
+  def perform()
```
acoffman, line 2: Don't actually have to change it, just a nitpicky FYI, but generally for zero argument methods, you'd omit the `()` but that's just a common idiom, not required/has no impact on anything.

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,20 @@
...
+  Sidekiq::Cron::Job.load_from_hash YAML.load_file(schedule_file)
+end 
+
+Sidekiq::Web.use(Rack::Auth::Basic) do |user, password|
+  # Protect against timing attacks:
+  # - See https://codahale.com/a-lesson-in-timing-attacks/
+  # - See https://thisdata.com/blog/timing-attacks-against-string-comparison/
+  # - Use & (do not use &&) so that it doesn't short circuit.
+  # - Use digests to stop length information leaking
+  ActiveSupport::SecurityUtils.secure_compare(user, ENV["SIDEKIQ_ADMIN_USER"]) &
```
acoffman, line 18: We may want to pull the admin user and password out into variables and then do something like we do for the oauth credentials in civic 

```
password = ENV['SIDEKIQ_ADMIN_USER'] || Rails.secrets.sidekiq_admin_user

```

(or whatever the exact syntax is) so that we can use env variables locally and the capistrano deployed secrets file in prod.

app/jobs/tsv_updater.rb
```diff
@@ -4,18 +4,14 @@
...
-      pre_grouper = PreGrouper.new
-      pre_grouper.perform
-      grouper = Grouper.new
-      grouper.perform(should_group_genes?, should_group_drugs?)
-      post_grouper = PostGrouper.new
-      post_grouper.perform(should_cleanup_gene_claims?)
-    ensure
-      reschedule if recurring
-    end
+  def perform()
```
acoffman, line 7: Just a silly nitpick/fyi, no need to actually change it as it doesn't impact anything, but idiomatically you'd omit the `()` for a method definition with no arguments.

config/scheduled_tasks.yml
```diff
@@ -0,0 +1,53 @@
+UpdateCgi:
+  cron: '0 0 1 * *'
```
acoffman, line 2: One thing that might be worth doing here that I tried to do with the civic scheduler is space the jobs out some so that there's not like one specific period where several of the jobs are all running at the same time and making the box slow. So maybe like if they're weekly some go on Saturday and some go on Tuesday or whatever.

PR #399, opened 2020-06-16
WIP: Add linkout to source's interaction claim pages

### LINKED ISSUE #304: link to actual source page for interaction claim
As per @obigriffith, it's not clear if or how we do this, currently. For each interaction claim, we want to be able to link directly to the actual source page from which we obtained information for the claim.

susannasiebert:
Closes #304 

I checked on some of the other online updaters but I couldn't find any source pages there that we could link out to (most times the interactions are just in a table on a gene/drug page instead of having individual pages).

Side note: With the API rate limiting of CIViC we will almost certainly run into problems with the CIViC importer. I'll make a separate issue to address that.


susannasiebert:
For other sources we should link to e.g. the page that has the table we import (so the link would be the same for all interaction claims from that source).

susannasiebert:
For the following sources I was unable to identify where the linkout should point to:
- [ ] Cancer Commons
- [ ] Clearity Foundation Biomarker
- [ ] Clearity Foundation Clinical Trial
- [ ] DoCM
- [ ] Guide To Pharmacology Interactions

For the importers/updaters I was able to update, I would also appreciate a closer look at the linkout URLs used to confirm that they are correct.

ahwagner:
- **Cancer Commons**: the site has changed quite a bit since the 2013 dataset we downloaded. I think we should just link out to the .tsv we collected this data from.
- **Clearity Foundation Biomarker**: same as above. If we have the .tsv, we should point to it.
- **Clearity Foundation Clinical Trial**: same as above.
- **DoCM**: just the homepage
- **Guide to Pharmacology Interactions**: link to the ligand page for the interaction. If we have the ligand id as an attribute for the corresponding interaction claim, we can link directly to the ligand page, e.g. for PAZOPANIB (id 5698):
https://www.guidetopharmacology.org/GRAC/LigandDisplayForward?ligandId=5698
Otherwise, we can hijack the search URL with the drug name:
https://www.guidetopharmacology.org/GRAC/LigandTextSearchForward?searchString=PAZOPANIB&searchCategories=all&order=rank&submitName=Search+the+database

susannasiebert:
For Cancer Commons and the two Clarity Foundation ones, we do have the TSVs in the data subrepo but that's not really publicly accessible, right? I haven't been able to find the originals on the sources' websites.

susannasiebert:
This PR is ready for final review.

PR #401, opened 2020-06-16
Redirect exact search matches to the gene/drug's interactions tab

### LINKED ISSUE #125: Exact matches pull up interactions tab on gene or drug view
When the user searches for only one gene or drug and we have an exact match for it in the database, the search result should be the interactions tab on that gene or drug.

susannasiebert:
Closes #125


susannasiebert:
Good catch. It did indeed not work for things like `INSULIN ZINC SUSP PURIFIED BEEF/PORK`. URL encoding the name fixed that issue.

PR #405, opened 2020-06-25
Allow a source to have multiple types

### LINKED ISSUE #365: Allow sources to have multiple source types
This will require merging of some sources, some view changes (e.g. source page) and importer updating for some sources.

acoffman:
This PR contains the database changes, model changes, and basic view changes to allow for this. It does leave some open questions to discuss in the meeting however:

- [ ] Which importers do we need to port forward (basically, what is maintained? Online ones almost certainly, but TSV too?)

- [ ] Now that we have this functionality, which sources do we want to leverage it?

- [ ] How do we want to display the icons/colors for a source that has multiple types? Right now it just picks one

closes  #365


susannasiebert:
Some sources to consider for this change:
- COSMIC is a good candidate for multiple source types (interaction as well as potentially druggable). If we do this we would want to add the "drug resistance" gene category to the gene claims.
- CIViC could be multiple source types as well, similar to COSMIC with the difference that we currently import all predictive evidence items (of certain rating, status, etc) but we would only add the "drug resistance" gene category for the genes involved in predictive evidence items that have a "resistance" clinical significance with evidence direction of "supports".
- Foundation One currently has two importers but the interaction source is no longer in the database and the source tsv could not be located. I think we just need to delete the interaction importer for this source.
- Guide to Pharmacology currently has two importers, one for interactions and one for potentially druggable genes. The each import from different TSVs so we could either keep them separate (but use the same source), or we could combine them and first import the gene tsv to create gene claims and assign gene categories and then import the interaction tsvs to import the interactions (and add additional aliases to the previously generated gene claims). The interaction source already has an online updater to so it would be ideal to add the gene import here as well to keep both up-to-date on the same schedule.

I think that's all of them that are both interaction and druggable gene categories. I think we will need to change the importers at the same time and reimport in order to get everything assigned correctly (since all the claims are linked to the source). I can take a stab at updating these importers to use this new paradigm.

ahwagner:
From discussion on call earlier:

Yes, COSMIC and CIViC as "drug resistance" claims sounds good, with the CIViC interactions filtered in the way specified.

CIViC as "clinically actionable" is also good, but should be limited to level A evidence.

Deleting the Foundation One importer seems like the only course left. +1.

GtP plans sound good.

ChEMBL merge sounds good.

Let's hold off on these changes until after we wrap up the most recent drug grouper and NDex export.

susannasiebert:
I finished COSMIC, CIViC, and GuideToPharmacology. The rest is still outstanding but I think we can merge what we have so far (after thorough review).

ahwagner (approved):

PR #408, opened 2020-06-30
Fix query score calculation

susannasiebert:
The reworked query score was missing the 1/ portion of the average
inverse promiscuity calculation.

### DIFF
--- a/app/presenters/interaction_search_results_presenter.rb
+++ b/app/presenters/interaction_search_results_presenter.rb
@@ -48,7 +48,7 @@ class InteractionSearchResultsPresenter
       promiscuity_count = DataModel::Interaction.where(identifier => result_interaction.send(identifier)).count
       pub_count = result_interaction.publications.count
       source_count = result_interaction.sources.count
-      all_promiscuity_counts = DataModel::Interaction.group(identifier).count.values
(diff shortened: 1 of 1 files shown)


app/presenters/interaction_search_results_presenter.rb
```diff
@@ -48,7 +48,7 @@ def scores_for_result(result, grouped_results)
       promiscuity_count = DataModel::Interaction.where(identifier => result_interaction.send(identifier)).count
       pub_count = result_interaction.publications.count
       source_count = result_interaction.sources.count
-      all_promiscuity_counts = DataModel::Interaction.group(identifier).count.values
+      all_promiscuity_counts = DataModel::Interaction.group(identifier).count.values.map{|x| 1/x}
```
acoffman, line 51: Can `x` ever be zero here?

susannasiebert: `group`, works only on things that are in the interactions table and then `count` will give counts for each `identifier` (`gene_id` or `drug_id`), e.g. how many interactions are there for each gene_id or drug_id that is involved in an interaction. So that shouldn't ever return 0 since it will only use gene ids and drugs ids that are actually part of an interaction.

acoffman: Makes sense

Issue #353, opened 2019-09-03
add scores to API results


ahwagner:
We currently have interaction scores on the website, but not in our interactions API. We should revise our API to return these as well.

susannasiebert:
The updated query score depends on the other results in the search space. So for an individual interaction, the score is not meaningful/can't be calculated.

susannasiebert:
The score does get returned on the api endpoint for interaction searches.

susannasiebert:
During today's meeting we discussed implementing an additional global interaction score to be displayed on a particular interaction in addition to the new query-specific interaction score.

PR #423, opened 2020-08-05
Add a global interaction score

### LINKED ISSUE #402: Implement global interaction score and add to API endpoint

susannasiebert:
Closes #402 and #353


susannasiebert:
It really isn't too bad since I pulled out the `known_drug_partners_per_gene` and `known_gene_partners_per drug` to be calculated once in `app/views/interaction_claims/_interactions_search_panel.html.haml` and getting passed along to the `interaction_score` method. It does execute each of those calculations twice, once per tab (`Unique Matches` and `Ambiguous or Unmatched`). Each of those queries takes less than a second though.

acoffman:
Cool, well that seems good enough for now then and if it ends up too slow we can come back and optimize.

susannasiebert:
@ahwagner I would appreciate it if you could double check that I got the calculation right.

PR #314, opened 2017-11-17
WIP: New PharmGKB importer

kcotto:

Issue #396, opened 2020-06-04
Add license field to sources


susannasiebert:
We should include this as a new field in the database and display it on the source page and return it in the API.

PR #425, opened 2020-08-06
New pharmgkb job and importer

kcotto:
Takes the place of #314. Now downloads zip containing interactions.tsv and imports into DGIdb.


app/jobs/update_pharmgkb.rb
```diff
@@ -0,0 +1,50 @@
+require 'net/http'
+require 'openssl'
+require 'zip'
+
+OpenSSL::SSL::VERIFY_PEER = OpenSSL::SSL::VERIFY_NONE
+
+class UpdatePharmgkb < TsvUpdater
+  def tempfile_name
+    ['interactions', '.tsv']
```
susannasiebert, line 9: I would suggest a unique name here that is not also used by any of the other app/job TsvUpdaters. This is just in case the two would ever run at the same time so that one updater doesn't overwrite the tempfile of another updater.

lib/genome/importers/pharmgkb/new_pharmgkb.rb
```diff
@@ -0,0 +1,94 @@
...
+    def create_new_source
+      @source ||= DataModel::Source.create(
+          {
+              base_url:          'http://www.pharmgkb.org',
+              site_url:          'http://www.pharmgkb.org/',
+              citation:          "From pharmacogenomic knowledge acquisition to clinical applications: the PharmGKB as a clinical pharmacogenomic biomarker resource. McDonagh EM, Whirl-Carrillo M, Garten Y, Altman RB, Klein TE. Biomark Med. 2011 Dec;5(6):795-806. PMID: 22103613",
+              source_db_version:  get_version,
+              source_type_id:    DataModel::SourceType.INTERACTION,
+              source_db_name:    'PharmGKB',
+              full_name:         'PharmGKB - The Pharmacogenomics Knowledgebase'
```
susannasiebert, line 36: Since this overlaps with #396, please add a `license_link` and `license_text` field to this hash.

lib/genome/importers/pharmgkb/new_pharmgkb.rb
```diff
@@ -0,0 +1,94 @@
...
+          gene_name = row['Entity2_name']
+          pharmgkb_gene_id = row['Entity2_id']
+          drug_claim = create_drug_claim(drug_name, drug_name, 'PharmGKB Drug Name')
+          create_drug_claim_alias(drug_claim, 'PharmGKB ID', pharmgkb_drug_id)
+          gene_claim = create_gene_claim(gene_name, 'PharmGKB Gene Name')
+          create_gene_claim_alias(gene_claim, pharmgkb_gene_id, 'PharmGKB ID')
+          interaction_claim = create_interaction_claim(gene_claim, drug_claim)
+          if row['PMIDs'].present?
+            add_interaction_claim_publications(interaction_claim, row['Source'])
+          end
```
susannasiebert, line 56: Please add a source entity linkout here (as well as in the else block below) using the `create_interaction_claim_link` method (see https://github.com/griffithlab/dgi-db/pull/399/files for usage examples).

app/jobs/update_pharmgkb.rb
kcotto: changed to `pharmgkb_interactions.tsv`

lib/genome/importers/pharmgkb/new_pharmgkb.rb
kcotto: Added both of these fields.

lib/genome/importers/pharmgkb/new_pharmgkb.rb
kcotto: Added interaction linkout.

app/jobs/update_pharmgkb.rb
```diff
@@ -0,0 +1,50 @@
...
+    Zip::InputStream.open(StringIO.new(response.body)) do |io|
+      while entry = io.get_next_entry
+        if entry.name == 'interactions.tsv'
+          tempfile.write(io.read.encode("ASCII-8BIT").force_encoding("utf-8"))
+        end
+      end
+    end
+  end
+
+  def next_update_time
```
susannasiebert, line 32: With the implementation of sidekiq, we no longer need this method. Instead you need to add an entry to `config/scheduled_tasks.yml`. The top list of entries in this file are the monthly tasks but with staggered start times.

kcotto: Removed `next_update_time` function and added task to update PharmGKB in `config/scheduled_tasks.yml`.

susannasiebert:
@kcotto I made the initial import for this updater and created a data dump for it. I also found a couple of bugs in the process. Please have a look at my changes and maybe check out the data dump as well to see if it looks good to you.

kcotto:
Data dump looks great besides one small issue with importing the PMIDs which I just pushed a change for.

susannasiebert:
I just pushed another commit with an updated data dump for the missing PharmGKB publications.

kcotto:
Just checked the latest data dump. Looks good!

PR #424, opened 2020-08-05
New data dump with import of IDG data

susannasiebert:
Closes #334 and #395 

~Merging is blocked by #415 and #398~

PR #430, opened 2020-08-07
Data dump with directionality added for interaction claim types

### LINKED ISSUE #347: Add directionality to interactions
Curate and add a simple "inhibits" / "activates" column on interaction claims and groups.

susannasiebert:
There are quite a few interactions with interaction types with conflicting directionalities. We should discuss if and how to resolve those.

Closes #347

### DIFF
--- a/app/presenters/interaction_presenter.rb
+++ b/app/presenters/interaction_presenter.rb
@@ -11,8 +11,8 @@ class InteractionPresenter < SimpleDelegator
   end
 
   def na_or_types
-    types = interaction.interaction_types.map{|t| t.type}.reject{|t| t == "n/a"}
-    types.count > 0 ? types : ["n/a"]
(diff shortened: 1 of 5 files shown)


susannasiebert:
Digging through the code, this is because the importer isn't assigning gene claim categories to the generated gene claims.  It's importing the category as gene claim attributes instead. We seem to be a bit inconsistent on whether or not we import categories as a gene claim attribute and/or as a gene claim category; sometimes we import as both, sometimes as only one or the other. I'm thinking this almost needs to be a cleanup issue.

susannasiebert:
@kcotto I made #434 to address the gene claim category issue. I would prefer to address the IDG problem when I fix all the other sources.

susannasiebert:
Merging is blocked by #424

PR #431, opened 2020-08-11
Drug Target Commons Importer

kcotto:
Importer for Drug Target Commons. Listed as WIP because I need to add to `config/scheduled_tasks.yml` but there are currently other PRs that also have that file edited.


lib/genome/importers/dtc/dtc.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def create_interaction_claims
+    CSV.foreach(file_path, :headers => true) do |row|
+      next if row['compound_id'].nil? || row['gene_names'].nil?
+      drug_id = row['compound_id']
+      drug_name = row['compound_name']
+      drug_claim = create_drug_claim(drug_id, drug_id, 'CHEMBL ID')
+      if drug_name.present?
+        create_drug_claim_alias(drug_claim, drug_name, 'DTC drug name')
+      gene_name = row['gene_names']
```
susannasiebert, line 51: The `gene_names` column looks like it can contain more than one gene (separated by `, `). So you'll want to split on that string and iterate over them and create one gene claim and interaction claim for each.

app/jobs/update_dtc.rb
```diff
@@ -0,0 +1,34 @@
...
+  def importer
+    Genome::Importers::Dtc::DtcImporter.new(tempfile)
+  end
+
+  def download_file
+    uri = URI('https://drugtargetcommons.fimm.fi/static/Excell_files/DTC_data.csv')
+    download = open(uri)
+    IO.copy(download, tempfile_name)
+  end
+end
```
susannasiebert, line 19: These `end`s here (lines 19-21 and line 34) are extra. You will also want to indent all of the `should` methods to be on the same level as the rest of the methods.

lib/genome/importers/dtc/dtc.rb
susannasiebert, line 51: lines 51-53 need to be indented one more level

susannasiebert:
> I think we can use the `ep_action_mode` for our interaction claim type (with some normalization). E.g. `inhibition` we would want to import as `inhibitor`. I would suggest using the `name_normalizer` method in the InteractionClaimType normalizer. You will need to add missing terms they use to the appropriate if-elsif-else conditional

The values in this column are: `activation`, `cytotoxocity`, `growth_inhibition`, `inverse_agonist`, and `inhibition`

lib/genome/importers/dtc/dtc.rb
kcotto: Isn't that field being separated by `,` going to be an issue since the file itself is comma delimited?

susannasiebert: Wherever a comma is used within a field, the field is encased by `""` in the plain text file so the CSV parser will know to interpret this as one field and ignore the commas within the field, e.g.:
```
CHEMBL474208,AAAHTCZONLHYSA-UHFFFAOYSA-N,,,"Q96DB2, Q9UBN7, Q92769, Q13547, Q9UKV0, Q9UQL6, Q8WUI4, P56524, O15379, Q969S8, Q9BY41",HISTONE DEACETYLASE,"HDAC1, HDAC10, HDAC11, HDAC2, HDAC3, HDAC4, HDAC5, HDAC6, HDAC7, HDAC8, HDAC9",,,19301902,EC50,=,1740,NM,,"",cell_based,"","","","",HeLa,,,,,,,Inhibition of HDAC in human HeLa cells assessed as induction of cellular histone H3 hyperacetylation,"Design of chimeric histone deacetylase- and tyrosine kinase-inhibitors: a series of imatinib hybrides as potent inhibitors of wild-type and mutant BCR-ABL, PDGF-Rbeta, and histone deacetylases.",J. Med. Chem.,PUBLICATION,
```
notice how `"Q96DB2, Q9UBN7, Q92769, Q13547, Q9UKV0, Q9UQL6, Q8WUI4, P56524, O15379, Q969S8, Q9BY41"` and `"HDAC1, HDAC10, HDAC11, HDAC2, HDAC3, HDAC4, HDAC5, HDAC6, HDAC7, HDAC8, HDAC9"` each are encased in double quotes so it will be parsed as a single field. This is controlled by the `quote_char` option in the CSV reader and is by default set to `"`. See https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html for more info.

lib/genome/importers/dtc/dtc.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def create_interaction_claims
+    CSV.foreach(file_path, :headers => true) do |row|
+      next if row['compound_id'].nil? || row['gene_names'].nil?
+      drug_id = row['compound_id']
+      drug_name = row['compound_name']
+      drug_claim = create_drug_claim(drug_id, drug_id, 'CHEMBL ID')
+      if drug_name.present?
+        create_drug_claim_alias(drug_claim, drug_name, 'DTC drug name')
+      if row['gene_names'].include?(',')
```
susannasiebert, line 51: This should work the way it is written here but you don't have to check for the existence of the `,` first. You can just split on the text either way and if there is no `,` in it, it will return the whole string as an array with the string as the only item in it and it will then "loop over" this array.
```
$ rails c
Loading development environment (Rails 6.0.3.2)
[1] pry(main)> 'test'.split(',')
=> ["test"]
```

lib/genome/importers/dtc/dtc.rb
```diff
@@ -0,0 +1,78 @@
...
+        end
+        else
+          gene_name = row['gene_names']
+          gene_claim = create_gene_claim(gene_name, 'DTC gene name')
+          interaction_claim = create_interaction_claim(gene_claim, drug_claim)
+          pmid = row['pubmed_id']
+          unless pmid.nil? || pmid == '' || pmid =~ /-/
+            create_interaction_claim_publication(interaction_claim, pmid)
+          end
+          mechanism = row['ep_action_mode']
```
susannasiebert, line 68: The mechanism part here only gets executed for lines with a single gene in it. I would suggest to either make the change suggested above (removing the if-else) or you will need to copy the code for the mechanism part into the if-clause as well.

lib/genome/importers/dtc/dtc.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def create_interaction_claims
+    CSV.foreach(file_path, :headers => true) do |row|
+      drug_id = row['compound_id']
+      drug_name = row['compound_name']
+      gene_name = row['gene_names']
+      pmid = row['pubmed_id']
+      mechanism = row['ep_action_mode']
+      unless drug_name.nil? || gene_name.nil?
+        drug_claim = create_drug_claim(gene_name, gene_name, 'DTC Drug Name')
```
susannasiebert, line 51: You are using the `gene_name` here instead of the `drug_name`

kcotto: I fixed this

Issue #393, opened 2020-06-04
Update PharmGKB and DrugBank information on Downloads page


kcotto:
Both of these sources are now under creative commons licenses. We should update this information on the downloads page (http://www.dgidb.org/downloads)

kcotto:
Look into whether any other sources fall into this category as well and update them on this page too.

PR #432, opened 2020-08-11
Remove pharmgkb text

kcotto:
Addresses #393. Removed text for PharmGKB since its license has changed. Left DrugBank since it still has a custom license.


susannasiebert:
We should check that the PharmGKB data is now included in the interaction tsv.

kcotto:
Removed code that excludes PharmGKB from tsv dump.

PR #433, opened 2020-08-12
Autogenerate the interaction types table

### LINKED ISSUE #358: autogenerate getting started types table
Sharon to migrate db to include definitions for terms.

Suzi to autogenerate the types table.

susannasiebert:
This PR creates a new page "Interaction Types" that contains the interaction types table previously found on the Getting Started page. It fill in the definitions/descriptions of all types in the database from the old table. Additional, this PR adds a new column on interaction claim types to capture the values from the Reference column of the table (stored in plain text with html link embedded).

Lastly, this PR renames the Getting Started page to About and removes mentions of tutorials and such, where I found them.

Closes #358


susannasiebert:
You might need to migrate your database first `rake db:migrate` and load the data dump associated with this PR.

susannasiebert:
We can definitely make a change like that. I just copied it straight from the old table.

text/faq.yml
```diff
@@ -25,13 +24,13 @@ faq:
...
 
-    - question: How can I learn more about how to use DGIdb?
-      anchor: learn
-      answer: > 
-        You can try reading: the <a href="/getting_started">'Getting Started'</a> section under the help menu,
-         the <a href="https://doi.org/10.1093/nar/gkx1143">manuscript</a>, and this FAQ.
-        Some sections also contain a 'Tour' that can be activated by pressing the 'Show Tour' button in the top left.
-        If all else fails, please <a href="mailto:help@dgidb.org">email us</a> with your questions.
+    #- question: How can I learn more about how to use DGIdb?
+    #  anchor: learn
```
acoffman, line 28: Is this intended to be commented rather than removed?

susannasiebert: I figured we might want to add this again if we ever add another set of tutorials.

susannasiebert:
identifiers should be structured like this: https://registry.identifiers.org/registry/pubmed

susannasiebert:
@kcotto I updated to PubMed references to match the identifiers.org format.

PR #443, opened 2020-08-20
Add sidekiq config file

susannasiebert:


lib/systemd/system/sidekiq.txt
```diff
@@ -0,0 +1,36 @@
+[Unit]
+Description=sidekiq
+# start us only once the network and logging subsystems and redis are available
+After=syslog.target network.target redis-server.service
+
+[Service]
+Type=simple
+WorkingDirectory=/var/www/myapp/current/
```
acoffman, line 8: we'll want to change that to be the actual dgidb dir instead of `myapp`

PR #450, opened 2020-08-24
OnlineUpdater for Drug Target Commons using their API

susannasiebert:
This is an alternative for #431. The CSV file used in #431 is over 2Gb and takes a very long time to download using the updater (it ran for an hour for me and still didn't finish downloading). Using the API would hopefully be faster. This is still a WIP. I'm currently running to importer to see how long it takes and to evaluate the data after importing.


susannasiebert:
I wonder if we can even dial up the number of entries requested at a time to something higher than 500 since there are 14million entries in total.

Issue #329, opened 2019-04-18
Add documentation on how filters work to DGIdb FAQ page


ahwagner:
Describe the design and use of the drug filters (fda approved, antineoplastic, immunotherapy) and gene filters (clinically actionable, druggable genome, drug resistance) in the web search interface.

PR #451, opened 2020-08-25
Filter documentation

kcotto:
Addresses #329


kcotto:
Added a link for both interaction searches to link to the FAQ page @susannasiebert

PR #457, opened 2020-08-27
Some additional improvements to the interaction grouper

susannasiebert:
This PR mainly pulls up the code from `add_attributes` into the `add_member` method so that it only processes the attributes for newly created interactions. If we don't group from scratch I also moved the transaction to not happen around the whole process but one transaction for each interaction claim. This will make it so that you don't loose all the progress already made if a job fails or has to be aborted.

I also added an eager load for related tables on interaction claims in the hope that that will result in less SQL queries having to be executed.

This code is untested but I wanted to put it out there to get other people's opinions.

### DIFF
--- a/lib/genome/groupers/interaction_grouper.rb
+++ b/lib/genome/groupers/interaction_grouper.rb
@@ -1,19 +1,17 @@
 module Genome
   module Groupers
     class InteractionGrouper
-      attr_reader :group_from_scratch
       def self.run(group_from_scratch=false)
-        @group_from_scratch = group_from_scratch
         if group_from_scratch
           ActiveRecord::Base.transaction do
             puts 'reset members'
             reset_members
             puts 'add members'
-            add_members
+            add_members(group_from_scratch)
           end
         else
           puts 'add members'
-          add_members
+          add_members(group_from_scratch)
         end
       end
 
@@ -23,9 +21,9 @@ module Genome
         DataModel::Interaction.destroy_all
       end
 
-      def self.add_members
+      def self.add_members(group_from_scratch)
         DataModel::InteractionClaim.eager_load(:interaction_claim_types, :source, :interaction_claim_attributes, :publications).joins(drug_claim: [:drug], gene_claim: [:gene]).where(interaction_id: nil).each do |interaction_claim|
-          if @group_from_scratch
+          if group_from_scratch
             add_member(interaction_claim)
           else
             ActiveRecord::Base.transaction do
--- a/lib/genome/groupers/interaction_grouper.rb
+++ b/lib/genome/groupers/interaction_grouper.rb
@@ -1,16 +1,19 @@
 module Genome
   module Groupers
     class InteractionGrouper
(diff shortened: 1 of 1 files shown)


lib/genome/groupers/interaction_grouper.rb
```diff
@@ -1,16 +1,19 @@
 module Genome
   module Groupers
     class InteractionGrouper
+      attr_reader :group_from_scratch
```
acoffman, line 4: This will create an instance level accessor for `group_from_scratch` but all the methods here are class methods (denoted by the `def self.method_name` so it will not be usable in that context. You can access `@group_from_scratch` directly or you could add the accessor to the metaclass

```
class << self
  attr_reader :group_from_scratch
end
```

or if that's all too weird we could just pass the value along from `run()` to each of the sub-methods or switch things to be normal instance methods and `new` up a grouper.

I think I prefer one of those last two options myself. I generally make things just static class methods when there's no internal state, but if we want to start storing internal state for a class it makes more sense to either not store the state and just propagate it along, or convert to a normal class with instance level methods and variables.

I think actually it will "work" as written because you only access `@run_from_scratch` later in the code rather than through the accessor method, but if someone tried to use that accessor it would break.

lib/genome/groupers/interaction_grouper.rb
```diff
@@ -21,49 +24,59 @@ def self.reset_members
...
       def self.add_members
-        DataModel::InteractionClaim.joins(drug_claim: [:drug], gene_claim: [:gene]).where(interaction_id: nil).each do |interaction_claim|
-          drug = interaction_claim.drug_claim.drug
-          gene = interaction_claim.gene_claim.gene
-          interaction = DataModel::Interaction.where(drug_id: drug.id, gene_id: gene.id).first_or_create
-          interaction_claim.interaction = interaction
-          interaction_claim.interaction_claim_types.each do |t|
-            unless interaction.interaction_types.include? t
-              interaction.interaction_types << t
+        DataModel::InteractionClaim.eager_load(:interaction_claim_types, :source, :interaction_claim_attributes, :publications).joins(drug_claim: [:drug], gene_claim: [:gene]).where(interaction_id: nil).each do |interaction_claim|
```
acoffman, line 27: This isn't a change from your PR per se but I'm worried now that we're eager loading everything if we run this from scratch, if loading _all_ the interaction claims at once might be too much memory? We could change this `each` into `in_batches` (https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-in_batches) without any further code changes to avoid this.

lib/genome/groupers/interaction_grouper.rb
susannasiebert: I changed this to pass along `group_from_scratch` to the sub-methods.

lib/genome/groupers/interaction_grouper.rb
susannasiebert: I made this change.

PR #460, opened 2020-08-27
New importer for the Human Protein Atlas

susannasiebert:
TODO:
- [x] import data
- [x] add tsv to the data subrepo

The TSV was created by going to this page (https://www.proteinatlas.org/search/protein_class%3APotential+drug+targets) and selecting the columns Gene, Gene synonym, Ensembl gene id, Gene description, Uniprot accession, and Protein class (from the Gene section) and clicking "Download TSV".


lib/genome/importers/human_protein_atlas/human_protein_atlas.rb
```diff
@@ -0,0 +1,70 @@
...
+          if categories.has_key? c
+            gene_category = DataModel::GeneClaimCategory.find_by(name: categories[c])
+            gene_claim.gene_claim_categories << gene_category unless gene_claim.gene_claim_categories.include? gene_category
+          end
+        end
+      end
+    end
+
+    def categories
+      {
```
acoffman, line 60: I don't know how many rows there are in this dataset, and either way the impact is likely minimal but this method will create and retunr a new dictionary every time its called, which appears to be 2x per protein class per row which could add up. Memoizing it with 

```
def categories
  @categories ||= { 
     ....dictionary contents
}
end
```

Will only create the hash the first call and then just return a reference to that one each time its called. If it runs satisfactorily fast currently, you may not have to care though,

susannasiebert: Done.

PR #461, opened 2020-08-28
Update GO importer to work with new API

susannasiebert:


susannasiebert:
The one part I wasn't able to reproduce is:
```
id = gene['id'].gsub("#{gene['nomenclature']}:", "")
nomenclature = "#{gene['nomenclature']} ID"
create_gene_claim_alias(gene_claim, id, nomenclature)
```
I think that primarily used to populate UniProtID gene claim aliases. I didn't see anything equivalent in the new API response but this value might now be part of the synonym column.

susannasiebert:
Do we want to run this importer manually and make a new data dump with it?

susannasiebert:
Turns out that even the new API had the same problem. I fixed it now and the newest data dump has the correct set of genes.

PR #462, opened 2020-08-30
CMI importer updated to new style

kcotto:
New importer for Caris. Uploaded the TSV which was created from pulling the genes under point mutations, copy number alterations, and fusions from this page: https://www.carislifesciences.com/molecular-profiling-technology/
[CMI_point_CNA_fusions.txt](https://github.com/griffithlab/dgi-db/files/5147556/CMI_point_CNA_fusions.txt)


lib/genome/importers/caris_molecular_intelligence/caris_molecular_intelligence.rb
```diff
@@ -0,0 +1,46 @@
...
+              license: 'Unknown',
+              license_link: 'https://www.carismolecularintelligence.com/contact-us/',
+          }
+      )
+    end
+
+    def create_gene_claims
+      CSV.foreach(file_path, :headers => true, :col_sep => "\t") do |row|
+        gene_claim = create_gene_claim(row["Gene"], 'Gene Symbol')
+        gene_claim.gene_claim_categories << 'CLINICALLY ACTIONABLE'
```
susannasiebert, line 42: We might want to consider adding a new method to the online_updater.rb to `create_gene_claim_category`. This could look similar to `create_interaction_claim_type` and we would get the the desired category object from the database first and check that it's not already assigned to the `gene_claim`. As long as the gene-category combinations in the input TSV are unique, this shouldn't be needed but it would still clean up things a bit.

kcotto:
To wait on merging until after #465

susannasiebert (approved):
+1

PR #464, opened 2020-09-01
Rework drug grouper

susannasiebert:
This PR implements a new drug grouper. Right now it is called NewDrugGrouper but eventually it should replace the existing DrugGrouper (and moved in its place). This grouper queries the API designed in https://github.com/cancervariants/therapy-normalization for terms (drug claim name, primary name, and aliases) and normalizes the return value to a single concept. The logic for this is outlined in https://docs.google.com/document/d/1wSFJEJGzuwZz-Rf_-Gvy48OikdYON2fhkG83Od61PJ0/edit#

This PR also makes some changes to the dgidb schema:
- Rename drugs chembl_id column to concept_id
- Remove drugs chembl_molecule_id column
- Drop the chembl_molecules and chembl_molecule_synonyms tables
- Remove the unique constraint on the drugs name column and add a unique constraint on name+concept_id instead. This was done because two chembl ids can have the same name.

TODO:
- [x] The current code relies on the API running locally. We either need to run the therapy normalization API on the dgidb boxes (if that's even possible) or spin up a box that runs it.
- [x] The current code does not import other_identifiers. It wasn't clear to me whether those should be imported as aliases or as attributes
- [x] The current code does not pull up drug claim attributes, sources, or types to the drug level. 
- [x] The current code does not handle filling in the fda_approved flag.
- [x] We need to remove the PubChem sources in the database (and we can then remove drugs as a source category). This also includes removing the importer files for these sources. We need to remove ChEMBL and PubChem drug claims. Instead there will be WikiData and ChEMBL sources (with drug as the source category). There is not direct link between drugs and sources ~so the counters need to be custom populated either by the drug grouper or by the post grouper by querying the concept_id field of the drug table~ (edit: this isn't possible either since the counts are drug_claims and drug_claims_in_groups).
- [x] Update all the views to use the new concept_id instead of chembl_id
- [x] Update the new grouper to optionally only operate on the drug claims of a specific source.
- [x] Remove old grouper and update existing online updaters to use new grouper.
- [x] Remove references to the ChemblMolecule and ChemblMoleculeSynonym tables
- [ ] Refactor the drug grouper test


susannasiebert:
This drug grouper should also address https://github.com/griffithlab/dgi-db/issues/428. The wikidata terms usually only have their first letter capitalized while the Chembl terms are all-caps. I think we should capitalize the drug name when creating a drug object.

ahwagner (approved):

PR #463, opened 2020-08-30
New Foundation One importer

kcotto:
Updated to new style of importer. TSV is attached [foundation_one_union.txt](https://github.com/griffithlab/dgi-db/files/5147568/foundation_one_union.txt) and was 
created from taking the union of CDx - https://assets.ctfassets.net/w98cd481qyp0/YqqKHaqQmFeqc5ueQk48w/0a34fcdaa3a71dbe460cdcb01cebe8ad/F1CDx_Technical_Specifications_072020.pdf
Liquid - https://assets.ctfassets.net/w98cd481qyp0/7IpxsdNOTHT8SbzYLd6MV7/828dcc78b1fb4750680723819615fb03/F1L_TechnicalInformation_10__1_.pdf
Heme - https://assets.ctfassets.net/w98cd481qyp0/42r1cTE8VR4137CaHrsaen/baf91080cb3d78a52ada10c6358fa130/FoundationOne_Heme_Technical_Specifications.pdf

PR #465, opened 2020-09-01
Create gene claim category

kcotto:
Created to address comments on #462 and #463

### DIFF
--- a/lib/genome/online_updater.rb
+++ b/lib/genome/online_updater.rb
@@ -26,7 +26,9 @@ module Genome
 
     def create_gene_claim_category(gene_claim, category)
       gene_category = DataModel::GeneClaimCategory.find_by(name: category)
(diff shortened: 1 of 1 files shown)


lib/genome/online_updater.rb
```diff
@@ -24,6 +24,11 @@ def create_gene_claim_attribute(gene_claim, name, value)
       ).first_or_create
     end
 
+    def create_gene_claim_category(gene_claim, category)
+      gene_category = DataModel::GeneClaimCategory.find_by(name: category)
```
kcotto, line 28: Unsure if this is correct or if it should be something like `gene_category = DataModel:: GeneClaimCategory.where(
        type: type.strip()
      ).first_or_create`

susannasiebert: I think this is fine if we think that our current set of categories is all that we want. Any new ones we decide on would need to be added manually. This would also prevent us from accidentally importing junk categories. You'll probably want to throw some sort of error here if gene_category is nil.

PR #470, opened 2020-09-04
New importer for COSMIC

susannasiebert:
This might be a good candidate for a source with multiple source types once https://github.com/griffithlab/dgi-db/pull/405 is merged. We could then add this source to the potentially druggable source type and add the gene claims to the DRUG RESISTANCE category.

TODO:
- [x] make data dump
- [x] add csv file to data subrepo


kcotto (approved):
+1

PR #475, opened 2020-09-08
Various bugfixes

susannasiebert:


app/models/data_model/gene_claim.rb
```diff
@@ -53,10 +53,11 @@ def original_data_source_url
           entrez_id = self.gene_claim_aliases.select{|a| a.nomenclature == 'CKB Entrez Id'}.first.alias
           base_url + entrez_id
         when 'GO'
-          entrez_id = self.gene_claim_aliases.select{|a| a.nomenclature == 'UniProtKB ID'}.first.alias
-          if entrez_id.nil?
+          uniprot_alias = self.gene_claim_aliases.select{|a| a.nomenclature == 'UniProtKB ID'}.first
```
acoffman, line 56: Doesn't really matter but you could do 
`self.gene_claim_aliases.find {|a| a.nomenclature == 'UniProtKB ID'}`
which would be the same thing

acoffman (approved):
+1

PR #476, opened 2020-09-08
Data dump with Tempus imported

susannasiebert:


kcotto:
This looks good but I notice that there's no license info for Tempus. Is this intentional?
Screen Shot 2020-09-08 at 4 32 44 PM

susannasiebert:
Good catch. I made a new data dump that adds the license information. I also updated the citation to point to the source publications.

PR #491, opened 2020-09-14
Set drugbank username and password env variables when calling the drugbank fetching script

susannasiebert:


acoffman (approved):
Looks good to me; we'll just need to make sure we set these in the secrets file.

PR #502, opened 2020-09-22
added news for v4.0

sharonfreshour:


susannasiebert (changes requested):
+1 for the existing text. I would suggest to add a link to the preprint as well.

sharonfreshour:
Added the link, thanks for the suggestion @susannasiebert

susannasiebert (approved):

PR #507, opened 2020-09-23
Fix unmatched terms list when searching for interactions by drug

### LINKED ISSUE #115: Non-matching drugs collapse to one entry
When doing a query on drugs, all drugs without matches in the database are returned together as a single, non-matched item (comma separated). This is also the case in the returned .tsv file.

@gspies, see if you can find out what's going on here.

susannasiebert:
Closes #115


acoffman (approved):
+1

Issue #516, opened 2020-09-24
Speed up drug grouping


susannasiebert:
Some ideas:
- Reuse the http connection instead of setting up a new one for every request
- Batch requesting (this requires some changes in the therapy normalzier)
- Store "unmatchable" terms locally in a persistent manner.

PR #518, opened 2020-09-24
Use a persistent http connection for querying the therapy normalization API

susannasiebert:
Needed for #516

### DIFF
--- a/lib/genome/groupers/drug_grouper.rb
+++ b/lib/genome/groupers/drug_grouper.rb
@@ -3,7 +3,7 @@ require 'net/https'
 module Genome
   module Groupers
     class DrugGrouper
-      attr_reader :term_to_matches_dict, :term_to_record_dict, :valid_chembl_ids, :invalid_chembl_ids, :chembl_source, :wikidata_source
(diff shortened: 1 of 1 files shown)


lib/genome/groupers/drug_grouper.rb
```diff
@@ -178,10 +183,7 @@ def normalizer_matches_for_term(term)
         if term_to_matches_dict.has_key? term
           return term_to_matches_dict[term]
         end
-        uri = URI.parse(normalizer_url).tap do |u|
-          u.query = URI.encode_www_form( { q: ERB::Util.url_encode(term) } )
-          u.port = 8000
-        end
+        uri.query = URI.encode_www_form( { q: ERB::Util.url_encode(term) } )
         res = Net::HTTP.get_response(uri)
```
acoffman, line 187: This might work but I'm not actually 100% sure to be honest. I see that you're setting a persistent connection in the `URI` construction at the beginning but I'm unsure if this way of invoking it will actually use that or not. Have you tested it out? If not, I can give it a go.

acoffman (approved):

PR #538, opened 2020-10-02
When Drug is a ChEMBL identifier, remove the leading `chembl:` (includes data dump)

### LINKED ISSUE #531: Gene name sometimes is Chembl ID
Screen Shot 2020-10-01 at 11 12 11 AM
Unsure if we would rather have it default to a different name or at the very least remove the `chembl:` part of the name.

susannasiebert:
Closes #531


kcotto (approved):
+1

PR #543, opened 2020-10-08
Fix Show All/Less/More buttons in the interaction search results

### LINKED ISSUE #357: Bug with "show all" in interaction search results
There are two things going on here:
- Show all will expand all lists, not just the one for the search result you click "show all for"
- The counts in the "Showing x out of y interactions" are wrong

susannasiebert:
Closes #357.

The original code wasn't picking any of the particular result tables and applying the button action to all of them at the same time and then counting the rows for all tables. This code now assigns a unique id for each table. The buttons have a `value` matching the id of the respective table and the button action applies the action to only the table with that id.


acoffman (approved):
+1

PR #544, opened 2020-10-08
Speed up interaction search

susannasiebert:
`known_gene_partners_per_drug` and `known_drug_partners_per_gene` were previously calculated once per search result match. This change calculates it ones per search.


acoffman (approved):
+1 

For some future version, we may want to refactor the counts into a some object that contains all that metadata so we're not passing in so many params to each partial, but this seems good for now, especially given that we may re-do the views entirely for the next major release.

PR #546, opened 2020-10-09
Clean up gene categories that no longer have a matching gene claim category after source reimport (includes data dump)

susannasiebert:
I noticed that we have some genes that have a mismatch in categories depending on whether you do a category search or look at the gene summary directly. The first view is based on the underlying gene claims' categories and the latter is based on the gene categories directly (see https://dgidb.org/categories_search_results?genes=TRRAP#_successful and https://dgidb.org/genes/TRRAP for an example). This PR updated the database utils to delete extra gene categories after source deletion as well as interaction and gene grouping. It also updated the data dump to delete existing extraneous gene categories.

This is similar to the work done in https://github.com/griffithlab/dgi-db/pull/537


acoffman (approved):
+1

PR #547, opened 2020-10-09
Make DrugBank data available in the TSVs

### LINKED ISSUE #541: Update data downloads page
1.) Add a note at the top about licensing information for each source being available and link to https://www.dgidb.org/sources page

2.) Can we now include DrugBank data in our data dump TSVs?  Now that they use Attribution-NonCommercial 4.0 International?

susannasiebert:
Closes #541


acoffman (approved):
+1

PR #575, opened 2021-04-13
load_local fix

### LINKED ISSUE #572: database load_local not working
We no longer host the dgidb data subrepo, and the database export is intended to be hosted from the production site.

We need to create an export job and update the `load_local` routine to download the database from prod.

ahwagner:
closes #572.

Live on [staging](staging.dgidb.org). Ready to deploy after review.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "data"]
-	path = data
-	url = git://git.genomedata.org/dgidb_data.git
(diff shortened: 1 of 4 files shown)


susannasiebert (approved):
+1

Looks good to me. I tested it locally and it worked without problems.

acoffman (approved):
+1

PR #576, opened 2021-04-26
Handle cases where the single search term isn't a term in our database

susannasiebert:
This is a bug that was introduced by https://github.com/dgidb/dgidb/pull/401. When searching for a drug/gene that is not in DGIdb it would attempt to redirect, which would fail. This update will not trigger a redirect in this cases.


acoffman (approved):
+1

